PR #2, opened 2019-10-17
Changing namespaces

chathurawidanage:


src/twisterx.cpp
```diff
@@ -2,24 +2,25 @@
...
-using namespace twister::comm;
 
-
-class MyWorker : public Worker {
+class MyWorker : public twisterx::worker::Worker {
 
     void execute(Config config) override {
         std::cout << "Starting TwisterX program ..." << std::endl;
 
         int size = 10;
```
supunkamburugamuve: As suggested by Amila, lets use int_32 etc

src/api/comm/packers/IntegerPacker.cpp
```diff
@@ -0,0 +1,14 @@
+#include <array>
+#include "DataPacker.h"
+
+namespace twister::comm::packers {
+    class IntegerPacker : public DataPacker<int> {
+
+        void pack_to_buffer(int data, Buffer buffer, PackerStatus status) {
```
KRVPerera: Think Buffer and PackerStatus should be passed as reference here?
`void pack_to_buffer(int data, Buffer **&** buffer, PackerStatus **&** status) {`

If we don't need to make accidental copies of Buffer object in any case we can ask compiler to delete the copy ctor, and Copy assignment ctors also. 
`
Buffer(const Buffer &bf) = delete;
Buffer& operator=(const Buffer&) = delete;
`

PR #4, opened 2019-10-19
Proper deletion for array initialization

KRVPerera:
- Fixed memory leak of Buffer class adding a destructor
- Changed config class map to a unordered_map
   Think this is valid hashmap for configurations
- added compiler flags to enable extra compiler warnings


CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ IF(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "DEBUG")
 ENDIF(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "DEBUG")
 
 SET(CMAKE_REQUIRED_FLAGS "-std=c++17")
+add_compile_options(-Wall -Wextra -W)
```
thejkane: Good catch.

PR #10, opened 2020-02-11
Changing the directory structure and adding initial java folders and build files

supunkamburugamuve:


java/.idea/libraries/Maven__net_bytebuddy_byte_buddy_1_9_7.xml
```diff
@@ -0,0 +1,13 @@
+<component name="libraryTable">
```
chathurawidanage: .idea folder can be excluded from vcs

PR #35, opened 2020-03-19
removing pybind11 dependencies and examples

vibhatha:


cpp/setup.py
```diff
@@ -1,69 +0,0 @@
-import os
-import re
-import sys
-import platform
-import subprocess
-
-from setuptools import setup, Extension
```
supunkamburugamuve: I think these classes should go inside the python/twisterx directory

vibhatha: Yes, this was removed. That is true.

PR #65, opened 2020-04-20
Initial Python API for table [Experimental]

vibhatha:
Joins, merge are not added.


cpp/src/twisterx/table.hpp
```diff
@@ -20,109 +20,112 @@ namespace twisterx {
...
+        }
+
+        static Status from_csv(const std::string &path, const char &delimiter, const std::string &uuid);
+
+    private:
+        /**
+         * Every table should have an unique id
+         */
+        std::string id_;
+    };
```
chathurawidanage, line 128: seems there is an indentation settings mismatch?

vibhatha: @chathurawidanage yes, I am adding the Google format.

vibhatha:
@chathurawidanage @supunkamburugamuve 

I did some code clean up.

PR #73, opened 2020-05-14
Build Improvement and Modifications to CPP

vibhatha:


vibhatha:
Added a single build script to install CPP, PYARROW and PYTHON. 
Also added flags to build the project in chosen build modes (release, debug)

cpp/src/twisterx/python/ctx/twisterx_context_wrap.cpp
```diff
@@ -0,0 +1,101 @@
+//
+// Created by vibhatha on 5/15/20.
+//
```
supunkamburugamuve, line 3: Lets remove created by comment

vibhatha: I will remove that.

vibhatha: removed

PR #77, opened 2020-05-22
Adding Google Colab Notebooks to README [EXPERIMENTAL]

vibhatha:


cpp/src/twisterx/table.hpp
```diff
@@ -109,7 +109,7 @@ class Table {
    * @param out arrow table
    * @return the status of the operation
    */
-  Status ToArrowTable(std::shared_ptr<arrow::Table> *out);
+  Status ToArrowTable(std::shared_ptr<arrow::Table> &out);
```
supunkamburugamuve, line 112: this should be const std::shared_ptr

PR #104, opened 2020-07-08
fixed csv config holder memory leak

chathurawidanage:


cpp/src/cylon/io/csv_read_config.cpp
```diff
@@ -51,11 +50,11 @@ CSVReadOptions CSVReadOptions::SkipRows(int32_t skip_rows) {
 }
 
 CSVReadOptions::CSVReadOptions() {
-  CSVReadOptions::holder = new CSVConfigHolder();
+  CSVReadOptions::holder = std::shared_ptr<void>(new CSVConfigHolder());
 }
 
 void *CSVReadOptions::GetHolder() const {
-  return CSVReadOptions::holder;
+  return CSVReadOptions::holder.get();
```
supunkamburugamuve, line 57: I think we sould not take out the pointer from a shared pointer

PR #105, opened 2020-07-08
Intersect subtract

nirandaperera:


cpp/src/examples/union_example.cpp
```diff
@@ -22,9 +22,13 @@
  * This example reads two csv files and does a union on them.
  */
 int main(int argc, char *argv[]) {
-  if (argc < 3) {
-    LOG(ERROR) << "There should be two arguments with paths to csv files";
-    return 1;
+  std::string path1, path2;
+  if (argc != 3) {
+    path1 = "/tmp/csv1.csv";
+    path2 = "/tmp/csv2.csv";
```
supunkamburugamuve, line 28: I think this can be confusing to the user as we are expecting two files with fixed names in the /tmp folder

nirandaperera: (y)

cpp/src/cylon/table.cpp
```diff
@@ -166,17 +159,44 @@ Status Table::Select(const std::function<bool(cylon::Row)> &selector, shared_ptr
   }
   return status;
 }
-Status Table::DistributedUnion(const shared_ptr<Table> &right, shared_ptr<Table> &out) {
+
+Status Table::DoSetOperation(SetOperation operation, const shared_ptr<Table> &right, shared_ptr<Table> &out) {
   std::string uuid = cylon::util::uuid::generate_uuid_v4();
-  cylon::Status status = cylon::DistributedUnion(ctx, this->id_, right->id_, uuid);
+  cylon::Status status = operation(this->ctx, this->id_, right->id_, uuid);
```
chathurawidanage, line 165: Where is this operation() function located?

nirandaperera: its a typedef for a function pointer 
https://github.com/cylondata/cylon/blob/d62e8634b462ba8a6d78fa1b59a3a08956562ba3/cpp/src/cylon/table.hpp#L198

chathurawidanage: Okay. Took some time to understand that. :)

What's the advantage of doing this? Even if directly call the source function here,

```c++
Status Table::Union(const shared_ptr<Table> &right, std::shared_ptr<Table> &out) {
  return DoSetOperation(&cylon::Union, right, out);
}
```

it will take just one line, right? Does this improve readability or maintainability?

nirandaperera: Yes, because we can reuse DoSetOperation method for Union, Subtract, Intersection, and their dist variants. Otherwise, we have to repeat the same piece of code in all of these places. 
https://github.com/cylondata/cylon/blob/d62e8634b462ba8a6d78fa1b59a3a08956562ba3/cpp/src/cylon/table.cpp#L172
Take a look at likes 172-194

chathurawidanage: Got it. This is nice! Thanks.

PR #137, opened 2020-08-21
Adding columns API to the cylon table 

nirandaperera:


cpp/CMakeLists.txt
```diff
@@ -39,7 +39,13 @@ set(CMAKE_CXX_FLAGS "-Wall -Wextra")
 string(TOUPPER "${CMAKE_BUILD_TYPE}" UPPERCASE_CMAKE_BUILD_TYPE)
 if (UPPERCASE_CMAKE_BUILD_TYPE MATCHES "DEBUG")
     message("Running on debug mode...")
-    set(CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_FLAGS_DEBUG} -g -fsanitize=address")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g")
+
+    # if running tests, bypass address sanitizer, because catch header has memory leaks
+    #    if (NOT CYLON_WITH_TEST)
```
supunkamburugamuve, line 45: Remove the commented code

cpp/src/cylon/arrow/arrow_types.hpp
```diff
@@ -28,11 +28,18 @@ namespace tarrow {
  * @param tType the cylon type
  * @return corresponding arrow type
  */
-std::shared_ptr<arrow::DataType> convertToArrowType(std::shared_ptr<DataType> tType,
+std::shared_ptr<arrow::DataType> convertToArrowType(const std::shared_ptr<DataType> &tType,
```
supunkamburugamuve, line 31: WE should validate width, percision when using. otherwise it will set to -1 which not sensible

cpp/src/cylon/column.hpp
```diff
@@ -19,42 +19,95 @@
...
 
+#include <iostream>
+
 namespace cylon {
 
 class Column {
  public:
-  Column(std::string id, std::shared_ptr<DataType> type) : id(std::move(id)), type(std::move(type)) {
+  Column(const std::string &id, const std::shared_ptr<DataType> &type,
+         const std::shared_ptr<arrow::ChunkedArray> &data_)
```
supunkamburugamuve, line 34: I think we should not have _ at the end of data parameter

cpp/src/cylon/column.hpp
```diff
@@ -19,42 +19,95 @@
...
+
+  std::shared_ptr<arrow::ChunkedArray> data_;   // pointer to the data array
+};
+
+template<typename T, typename = typename std::enable_if<std::is_arithmetic<T>::value, T>::type>
+class VectorColumn : public Column {
+ public:
+  VectorColumn(const std::string &id,
+               const std::shared_ptr<DataType> &type,
+               const std::shared_ptr<std::vector<T>> &data_vector) :
```
supunkamburugamuve, line 85: Who owns this vector? Wha will happen if this vector goes out of scope and its memory is freed?

PR #171, opened 2020-09-30
adding table functions for slicing, indexing, schema, column-names

vibhatha:


python/pycylon/data/table.pyx
```diff
@@ -386,3 +386,14 @@ cdef class Table:
         py_arrow_table = pyarrow_wrap_table(table)
         return py_arrow_table.schema
 
+    def __getitem__(self, key):
+        table = to_pyarrow_table(self.id.encode())
+        py_arrow_table = pyarrow_wrap_table(table)
+        ctx = CylonContext(config=None)
```
nirandaperera, line 392: shouldn't we take ctx from the table itself? this reinitialization would be problematic, isn't it?

vibhatha: at the moment we don't support the wrapping and unwrapping C++ vs Python objects. We need to get that to do it. 
I am still exploring how to do that.

vibhatha: There is an easy fix in Python class to get the context when it is being built. That is the straightforward way to do it.

vibhatha: I checked this. Without Python CPP Object Wrapping Unwrapping, this won't be feasible at the moment.

vibhatha: Added an issue. 
https://github.com/cylondata/cylon/issues/172
Need to work on this separately for a major refactor.

vibhatha: @supunkamburugamuve @nirandaperera for this PR, the context creation issue was fixed.

nirandaperera (approved):

PR #187, opened 2020-10-15
PyCylon Refactor

vibhatha:


vibhatha:
@supunkamburugamuve @nirandaperera 
I didn't do rebase. Please squash when merging. 
Table.hpp refactor and func ptr callables, other code improvements will be added to a different PR. 
The aggregation will be added to the following PR.

python/pycylon/data/table.pyx
```diff
@@ -213,14 +248,112 @@ cdef class Table:
...
+        elif ra_op_name == 'distributed_union':
+            status = CTable.DistributedUnion(self.table_shd_ptr, right, output)
+        elif ra_op_name == 'intersect':
+            status = CTable.Intersect(self.table_shd_ptr, right, output)
+        elif ra_op_name == 'distributed_intersect':
+            status = CTable.DistributedIntersect(self.table_shd_ptr, right, output)
+        elif ra_op_name == 'subtract':
+            status = CTable.Subtract(self.table_shd_ptr, right, output)
+        elif ra_op_name == 'distributed_subtract':
+            status = CTable.DistributedSubtract(self.table_shd_ptr, right, output)
```
nirandaperera, line 349: add a final else here. I know this is an internal function, but if someone later decides to add a method, it would be helpful to debug.

python/pycylon/data/table.pyx
```diff
@@ -213,14 +248,112 @@ cdef class Table:
...
+        else:
+            raise ValueError(f"{op_name} operation failed!")
+
+    cdef _get_ra_response(self, table, ra_op_name):
+        cdef shared_ptr[CTable] output
+        cdef shared_ptr[CTable] right = pycylon_unwrap_table(table)
+        cdef CStatus status
+        # TODO: add callable for Cython functions via FPointers
+
+        if ra_op_name == 'union':
```
nirandaperera, line 338: As I said earlier, I would still prefer a switch case with an enum here. or better yet, use a function pointer (I am not sure how that could be done in cython though). Latter would eliminate the switch case, and it would be elegant IMO.

python/pycylon/data/table.pyx
```diff
@@ -213,14 +248,112 @@ cdef class Table:
...
+            status = CTable.DistributedIntersect(self.table_shd_ptr, right, output)
+        elif ra_op_name == 'subtract':
+            status = CTable.Subtract(self.table_shd_ptr, right, output)
+        elif ra_op_name == 'distributed_subtract':
+            status = CTable.DistributedSubtract(self.table_shd_ptr, right, output)
+
+        if status.is_ok():
+            return pycylon_wrap_table(output)
+        else:
+            raise ValueError(f"{ra_op_name} operation failed!")
```
nirandaperera, line 354: add the status msg in the exception

python/pycylon/data/table.pyx
```diff
@@ -259,156 +384,194 @@ cdef class Table:
...
+            if isinstance(columns[0], int) or isinstance(columns[0], str):
+                for column in columns:
+                    if isinstance(column, str):
+                        column = self._resolve_column_index_from_column_name(column)
+                    c_columns.push_back(column)
+                status = self.table_shd_ptr.get().Project(c_columns, output)
+                if status.is_ok():
+                    return pycylon_wrap_table(output)
+                else:
+                    raise ValueError("Project operation failed!")
```
nirandaperera, line 461: add the status msg in the exception

python/pycylon/data/table.pyx
```diff
@@ -259,156 +384,194 @@ cdef class Table:
...
-        return py_arrow_table
+        cdef shared_ptr[CCylonContext] ctx = pycylon_unwrap_context(context)
+        cdef shared_ptr[CArrowTable] arw_table = pyarrow_unwrap_table(pyarrow_table)
+        cdef shared_ptr[CTable] cn_table
+        cdef CStatus status = CTable.FromArrowTable(ctx, arw_table, &cn_table)
+
+        if status.is_ok():
+            return pycylon_wrap_table(cn_table)
+        else:
+            raise Exception("Table couldn't be created from PyArrow Table")
```
nirandaperera, line 483: add the status msg in the exception

python/pycylon/data/table.pyx
```diff
@@ -259,156 +384,194 @@ cdef class Table:
...
+        '''
+        cdef shared_ptr[CCylonContext] ctx = pycylon_unwrap_context(context)
+        cdef string cpath = path.encode()
+        cdef CCSVReadOptions c_csv_read_options = pycylon_unwrap_csv_read_options(csv_read_options)
+        cdef shared_ptr[CTable] cn_table
+        cdef CStatus status = CTable.FromCSV(ctx, cpath, cn_table, c_csv_read_options)
+        if status.is_ok():
+            return pycylon_wrap_table(cn_table)
+        else:
+            raise Exception("Table couldn't be created from CSV")
```
nirandaperera, line 501: add the status msg in the exception

python/pycylon/data/table.pyx
```diff
@@ -259,156 +384,194 @@ cdef class Table:
...
+         In the conversion we stack each column in the Table and create a numpy array.
+         For Heterogeneous Tables, use the generated array with Caution.
+         :param order: numpy array order. 'F': Fortran Style F_Contiguous or 'C' C Style C_Contiguous
+         :return: ndarray
+         """
         ar_lst = []
         _dtype = None
-        for col in py_arrow_table.columns:
+        for col in self.to_arrow().columns:
             npr = col.chunks[0].to_numpy()
```
nirandaperera, line 575: shall we do a merge chunks operation here, before converting to np?

python/pycylon/data/table.pyx
```diff
@@ -418,31 +581,58 @@ cdef class Table:
...
-        py_arrow_table = pyarrow_wrap_table(table)
-        return py_arrow_table.column_names
+        """
+        Produces column names from PyCylon Table
+        @return: list
+        """
+        return self.to_arrow().column_names
 
     @property
     def schema(self):
```
nirandaperera, line 624: I feel like we'd have to distinguish that this schema returned is an `arrow schema`. @supunkamburugamuve WDYT?

python/test/test_table_properties.py
```diff
@@ -1,30 +1,44 @@
...
-print(tb[2:5].to_pandas())
-
-print(tb[5].to_pandas())
-
-print(tb[7].to_pandas())
-
-tb.show_by_range(0, 4, 0, 4)
-
-print(tb[0:5].to_pandas())
+##
```
nirandaperera, line 1: if this test file is not required, let's remove it.

nirandaperera (changes requested):
Added some comments. BTW pls rebase and squash the commits into a single commit.

python/pycylon/data/table.pyx
vibhatha: why?

python/test/test_table_properties.py
vibhatha: no required. need to test the magic functions later.

vibhatha:
> Added some comments. BTW pls rebase and squash the commits into a single commit.

can't we do squash when merging?

python/pycylon/data/table.pyx
nirandaperera: it could be a chunked array and if we only convert the 0 chunk, we'll loose data, right?

nirandaperera:
> > Added some comments. BTW pls rebase and squash the commits into a single commit.
> 
> can't we do squash when merging?

Can do. You'll have to merge PR yourself at the end then. :-)

python/pycylon/__init__.py
```diff
@@ -12,6 +12,32 @@
...
+
+int = np.int
+int8 = np.int8
+int16 = np.int16
+int32 = np.int32
+
+uint = np.uint
+uint8 = np.uint8
+uint16 = np.uint16
+uint32 = np.uint32
```
supunkamburugamuve, line 28: is it a good idea to do this?

vibhatha: I will remove it. This was just for a test.

python/pycylon/data/table.pyx
```diff
@@ -213,14 +248,112 @@ cdef class Table:
...
+                                             right_column_index, CJoinAlgorithm.CSORT)
+            elif join_type == PJoinType.RIGHT.value:
+                self.jcPtr = new CJoinConfig(CJoinType.CRIGHT, left_column_index,
+                                             right_column_index, CJoinAlgorithm.CSORT)
+            elif join_type == PJoinType.OUTER.value:
+                self.jcPtr = new CJoinConfig(CJoinType.COUTER, left_column_index,
+                                             right_column_index, CJoinAlgorithm.CSORT)
+            else:
+                raise ValueError("Unsupported Join Type {}".format(join_type))
+        else:
```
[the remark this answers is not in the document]
vibhatha: I get your point. There is a constructor in JoinConfig: https://github.com/cylondata/cylon/blob/4a4720e0233d3e67e67cff7dbc112e4d366d9fac/cpp/src/cylon/join/join_config.hpp#L38

This one accepts all but an algorithm.

python/pycylon/data/table.pyx
vibhatha: could you elaborate a bit?

python/pycylon/data/table.pyx
vibhatha: Maybe we need to add our own Schema class in CPP or I can do it easily in Python itself. 
What should be the best?

supunkamburugamuve: If this is arrow schema, do we need to return it? Someone can get the arrow table and from there get the arrow schema?

python/pycylon/data/table.pyx
supunkamburugamuve: I think what Niranda tries to say is, if the table has chunked arrays (read from csv methods etc) this will result in we missing the rest of the data in the other chunks

python/pycylon/data/table.pyx
vibhatha: Agree. But our table also need a schema? We shouldn't encourage them to use arrow stuff in and out.

python/pycylon/data/table.pyx
vibhatha: I see, so once a table is generated after an RA op or just by loading. Does it have multiple chunks? 
Or is this like a validation?

supunkamburugamuve: It can have multiple chunks

nirandaperera: well, we can't guarantee that the underlying table only has a single chunk in each of its columns. You can use this method. 
http://arrow.apache.org/docs/python/generated/pyarrow.Table.html?highlight=combine%20chunks#pyarrow.Table.combine_chunks

python/pycylon/data/table.pyx
supunkamburugamuve: I think if it is a sepearate schema lets do it in another pull request. Lets remove this method for now

nirandaperera: @supunkamburugamuve WDYT?

python/pycylon/data/table.pyx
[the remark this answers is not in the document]
nirandaperera: Yeah. But it's already handled here
https://github.com/cylondata/cylon/blob/d3d57a028b6c99f1186edc2c37409ba5193866d7/python/pycylon/data/table.pyx#L264
So, I don't think we will need that constructor for python TBH.

python/pycylon/data/table.pyx
vibhatha:
```for col in self.to_arrow().combine_chunks().columns:
            npr = col.chunks[0].to_numpy()
```
This should work?

nirandaperera: yep. that should work

supunkamburugamuve (approved):

PR #189, opened 2020-10-16
Adding GroupBy and Aggregate

vibhatha:
Not sure how the old commits came up with this :/ (those were already merged)

Also, Groupby operation needs to be added without an aggregation operation. 
This need to be implemented in CPP side. Since that is not available. 
The Python API capability is limited when providing an API like Pandas groupby.


cpp/src/cylon/compute/aggregates.cpp
```diff
@@ -253,6 +252,192 @@ cylon::Status Max(const std::shared_ptr<cylon::Table> &table, int32_t col_idx, s
   return MinMax<1>(table, col_idx, output);
 }
 
+template<typename arrow_type, typename = typename std::enable_if<arrow::is_number_type<arrow_type>::value
```
nirandaperera, line 255: lets use upper case for types. May be ARROW_TYPE or ARROW_T

cpp/src/cylon/compute/aggregates.cpp
```diff
@@ -253,6 +252,192 @@ cylon::Status Max(const std::shared_ptr<cylon::Table> &table, int32_t col_idx, s
...
+
+  std::vector<std::shared_ptr<arrow::Field>> fields;
+  std::shared_ptr<arrow::Field> selected_filed = arw_table->schema()->field(col_idx);
+
+  fields.push_back(std::make_shared<arrow::Field>(selected_filed->name(),
+                                                  selected_filed->type(),
+                                                  selected_filed->nullable()));
+
+  auto schema = arrow::schema(fields);
+  auto out_a_table = arrow::Table::Make(schema, {out_vectors});
```
nirandaperera, line 295: I think this could be simplified like this (since there's only one element in the every vector)
```c++
 auto schema = arrow::schema({arw_table->schema()->field(col_idx)});
 auto out_a_table = arrow::Table::Make(schema, {out_index});
```

I don't think a separate array vector and a field vector is needed in this situation.

cpp/src/cylon/compute/aggregates.cpp
```diff
@@ -253,6 +252,192 @@ cylon::Status Max(const std::shared_ptr<cylon::Table> &table, int32_t col_idx, s
...
+
+  if (status.is_ok()) {
+    return arrow::Status().OK();
+  } else {
+    return arrow::Status::ExecutionError("Table resolution failed in aggregation");
+  }
+}
+
+typedef arrow::Status
+(*ResolveTableFromScalarFptr)(arrow::MemoryPool *pool,
```
nirandaperera, line 307: why don't we call the function itself, rather than returning the function ptr? I think that will simplify a lot of code in the useages. 
ex: for bool, we can simply call 
```c++
return ResolveTableFromScalar<arrow::BooleanType>(pool, ctx, col_idx, result, input, output);
```

cpp/src/cylon/compute/aggregates.cpp
```diff
@@ -253,6 +252,192 @@ cylon::Status Max(const std::shared_ptr<cylon::Table> &table, int32_t col_idx, s
...
+
+cylon::Status Min(const std::shared_ptr<cylon::Table> &table,
+                           int32_t col_idx,
+                           std::shared_ptr<cylon::Table> &output) {
+  cylon::Status status;
+  std::shared_ptr<cylon::compute::Result> result;
+  std::shared_ptr<cylon::CylonContext> ctx = table->GetContext();
+
+  if ((status = cylon::compute::Min(table, col_idx, &result)).is_ok()) {
+    ResolveTableFromScalarFptr fptr = CreateTableFromScalar(result);
```
nirandaperera, line 419: check the above comment. Then, lines 419-421 would be a single function call. there's no need for a function pointer IMO.

nirandaperera: BTW we'll have to add another status check actually, to check if table creation was successful or not

cpp/src/cylon/compute/aggregates.cpp
```diff
@@ -253,6 +252,192 @@ cylon::Status Max(const std::shared_ptr<cylon::Table> &table, int32_t col_idx, s
   return MinMax<1>(table, col_idx, output);
 }
 
+template<typename arrow_type, typename = typename std::enable_if<arrow::is_number_type<arrow_type>::value
+                                                                     | arrow::is_boolean_type<arrow_type>::value>::type>
+arrow::Status ResolveTableFromScalar(arrow::MemoryPool *pool,
+                                     std::shared_ptr<cylon::CylonContext> &ctx,
+                                     int32_t col_idx,
+                                     std::shared_ptr<cylon::compute::Result> &result,
+                                     std::shared_ptr<cylon::Table> input,
```
nirandaperera, line 261: I think both result and input can be made `const &`

cpp/src/examples/compute_example.cpp
```diff
@@ -49,42 +50,69 @@ int main() {
...
+  if ((status = cylon::compute::Max(table, agg_index, &result)).is_ok()) {
     const std::shared_ptr<arrow::DoubleScalar>
         &aa = std::static_pointer_cast<arrow::DoubleScalar>(result->GetResult().scalar());
     std::cout << "max " << aa->value << " " << status.get_code() << std::endl;
   } else {
     std::cout << "max failed! " << status.get_msg() << std::endl;
   }
 
+  // Adding Python funcs
+  std::shared_ptr<cylon::Table> output;
```
nirandaperera, line 90: Let's add these parts to the tests as well. 
https://github.com/cylondata/cylon/blob/master/cpp/test/aggregate_test.cpp

python/test/test_table_from_numpy.py
```diff
@@ -84,10 +84,84 @@
...
+print(op1.name)
+
+dict3 = {'col1': [1, 2, 3, 4, 5, 1, 3, 6, 8, 1, 9, 10], 'col2': [2, 4, 0, 1, 5, 6, 8, 1, 3, 4, 0,
+                                                                 1]}
+
+cn_tb14: cn.Table = cn.Table.from_pydict(ctx, dict3)
+
+cn_tb14.groupby(0, [0], [AggregationOp.COUNT])
+
+cn_tb14.show()
```
nirandaperera, line 146: lets use asserts to actually check for the correct value

nirandaperera (changes requested):

nirandaperera:
@vibhatha coming to think of the aggre

nirandaperera:
sorry I mistakenly closed this PR! :grin:

vibhatha:
Also reminding this: Groupby operation needs to be added without an aggregation operation.
This need to be implemented in CPP side. Since that is not available.
The Python API capability is limited when providing an API like Pandas groupby.

We may need to change how we do GroupBy and return an intermediate output. I think that is needed.

python/test/test_table_from_numpy.py
vibhatha: the test bench will be separately added for all Python use cases.

nirandaperera (approved):

nirandaperera:
> Also reminding this: Groupby operation needs to be added without an aggregation operation.
> This need to be implemented in CPP side. Since that is not available.
> The Python API capability is limited when providing an API like Pandas groupby.
> 
> We may need to change how we do GroupBy and return an intermediate output. I think that is needed.

I agree. Like I said previously, what pandas groupby returns is the indices of each group, and then we can apply an aggregation on each group. That is not so efficient for distributed operations because we would have to do a shuffle to get all the index rows in one place.

PR #215, opened 2020-10-28
CYLON_PARQUET flag

Sandeepa1995:
add cmake flag CYLON_PARQUET to build parquet related files (default OFF)


cpp/src/cylon/table.hpp
```diff
@@ -204,6 +198,17 @@ class Table {
    */
   std::vector<std::shared_ptr<cylon::Column>> GetColumns() const;
 
+#ifdef BUILD_CYLON_PARQUET
+  /**
+  * Write the table as a parquet file
+  * @param path file path
+  * @return the status of the operation
+  */
```
supunkamburugamuve, line 206: Shall we put all the parquet related methods under one check?

supunkamburugamuve:
We need to add this flag and enable parquet in the travis build. Look at the .travis.yaml file in the root folder.

cpp/src/cylon/table.hpp
Sandeepa1995: In a single #Ifdef ... #endif segment in the file? The issue is that WriteParquet is a part of Table's public functions while the reads are not. Or do you mean to put all the parquet related code into separate files and only check in table_api? (like table_api calling from parquet.hpp rather than table.hpp)

supunkamburugamuve: Got it. I think I missed that part

supunkamburugamuve (approved):

Sandeepa1995:
> We need to add this flag and enable parquet in the travis build. Look at the .travis.yaml file in the root folder.

Ah, okay. Should I add that to the current cpp build directly, create a separate flag and add to the build or create a separate build entirely?

supunkamburugamuve:
How about we add a flag to build.sh that can specify any number of CMAKE flags? ./build.sh --cmake_flags "-D" ?

PR #223, opened 2020-10-31
Python api v3

vibhatha:


vibhatha:
Testing Github workflow with Before Arrow Upgrade

python/examples/science_benchcmarks/uno/data_processing.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def load_aggregated_single_response_pandas(target='AUC', min_r2_fit=0.3, max_ec50_se=3.0,
+                                           combo_format=True,
+                                           rename=False):
+    url = "https://ftp.mcs.anl.gov/pub/candle/public/benchmarks/Pilot1/combo/combined_single_response_agg"
+    output_combined_single_response = \
+        "/home/vibhatha/data/uno/Pilot1/workload_1/combined_single_response_agg"
+
+    if not os.path.exists(output_combined_single_response):
```
supunkamburugamuve, line 21: Shall we move these to another repository?

vibhatha: Sure let’s do that

python/pycylon/index.py
```diff
@@ -0,0 +1,210 @@
...
+    @property
+    def index_values(self):
+        return self._index
+
+
+def range_calculator(rg: range):
+    start = rg.start
+    step = rg.step
+    stop = rg.stop
+    if step != 1:
```
nirandaperera, line 115: I think we can simplify this to 
```python 
  return math.ceil((stop - start)/step)
```
IMO we need to avoid conditional clauses as much as possible

python/pycylon/index.py
```diff
@@ -0,0 +1,210 @@
...
+            return length
+        else:
+            return length + 1
+    else:
+        return stop - start
+
+
+def _is_index_and_range_validity(table, index_range):
+    if isinstance(index_range, range):
+        if range_calculator(index_range) == table.row_count:
```
nirandaperera, line 128: this can be 
```python
return range_calculator(index_range) == table.row_count
```

nirandaperera (changes requested):

supunkamburugamuve (approved):

PR #252, opened 2020-12-17
Partition python

vibhatha:


nirandaperera (approved):
shall we keep the join_in_mem example intact and make a new example for the shuffle?

PR #273, opened 2021-01-09
Cpp indexing

vibhatha:


cpp/src/cylon/indexing/index.cpp
```diff
@@ -0,0 +1,28 @@
+#include "index.hpp"
+
+namespace cylon {
+
+std::unique_ptr<IndexKernel> CreateHashIndexKernel(std::shared_ptr<arrow::Table> input_table, int index_column) {
+  switch (input_table->column(index_column)->chunk(0)->type()->id()) {
+
+    case arrow::Type::NA:return nullptr;
```
nirandaperera, line 8: shall we implement the others as well?

cpp/src/cylon/indexing/index.cpp
```diff
@@ -0,0 +1,28 @@
...
+    case arrow::Type::INT64: return std::make_unique<Int64HashIndexKernel>();
+    case arrow::Type::HALF_FLOAT:return nullptr;
+    case arrow::Type::FLOAT:return nullptr;
+    case arrow::Type::DOUBLE:return nullptr;
+    case arrow::Type::STRING:return nullptr;
+    case arrow::Type::BINARY:return nullptr;
+    default: return nullptr;
+  }
+  //TODO : returning nullptr issue
+  return std::unique_ptr<IndexKernel>();
```
nirandaperera, line 26: what would be this default index kernel?

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+  int size_;
+  int col_id_;
+};
+
+template<class ARROW_T, typename CTYPE = typename ARROW_T::c_type>
+class Index : public BaseIndex {
+ public:
+  using ARROW_ARRAY_TYPE = typename arrow::TypeTraits<ARROW_T>::ArrayType;
+  using MMAP_TYPE = typename std::unordered_multimap<CTYPE, int64_t>;
+  explicit Index(int col_ids, int size, MMAP_TYPE map) : BaseIndex(col_ids, size) {
```
nirandaperera, line 52: we don't need explicit here

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+    const arrow::Datum filter_indices(out_idx);
+    arrow::Result<arrow::Datum> result = arrow::compute::Take(input_table, filter_indices, arrow::compute::TakeOptions::Defaults(), &fn_ctx);
+    RETURN_CYLON_STATUS_IF_ARROW_FAILED(result.status());
+    output = result.ValueOrDie().table();
+    return Status::OK();
+  }
+
+  std::shared_ptr<arrow::Array> GetIndex() override {
+
+    using ARROW_ARRAY_T = typename arrow::TypeTraits<ARROW_T>::ArrayType;
```
nirandaperera, line 85: this is already defined in #L50

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+    std::vector<CTYPE> vec(GetSize(), 1);
+    LOG(INFO) << "Get Index :: " << GetSize();
+
+    for(const auto &x: map_){
+      std::cout<< x.first <<":"<< x.second << std::endl;
+      vec[x.second] = x.first;
+    }
+
+    builders.AppendValues(vec);
+    arrow_status = builders.Finish(&index_array);
```
nirandaperera, line 105: shall we use the arrow::Result API here?

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+class Index : public BaseIndex {
+ public:
+  using ARROW_ARRAY_TYPE = typename arrow::TypeTraits<ARROW_T>::ArrayType;
+  using MMAP_TYPE = typename std::unordered_multimap<CTYPE, int64_t>;
+  explicit Index(int col_ids, int size, MMAP_TYPE map) : BaseIndex(col_ids, size) {
+    map_ = map;
+  };
+
+  Status Find(void *search_param,
+              std::shared_ptr<cylon::CylonContext> &ctx,
```
nirandaperera, line 57: shall we accept an arrow::MemoryPool here. I think there's no need to get the ctx

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+  std::shared_ptr<BaseIndex> BuildIndex(std::shared_ptr<arrow::Table> &input_table,
+                                        const int index_column,
+                                        bool index_drop,
+                                        std::shared_ptr<arrow::Table> &output_table) override {
+
+    const std::shared_ptr<arrow::Array> &idx_column = input_table->column(index_column)->chunk(0);
+    MMAP_TYPE out_umm_ptr = MMAP_TYPE(idx_column->length());
+    auto reader0 = std::static_pointer_cast<ARROW_ARRAY_TYPE>(idx_column);
+    for (int64_t i = 0; i < reader0->length(); i++) {
+      auto val = reader0->GetView(i);
```
nirandaperera, line 151: let's not use get view here. Let's use reader0->Value()

cpp/src/cylon/indexing/index.hpp
nirandaperera: This is going to be a bit of a problem @vibhatha. this map will be copied to Index class in every construction. we'll have to use a shared_ptr of a multimap here.

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+      out_umm_ptr.insert(std::make_pair(val, i));
+    }
+    auto index = std::make_shared<Index<ARROW_T, CTYPE>>(index_column, input_table->num_rows(), out_umm_ptr);
```
nirandaperera, line 154: out_umm_ptr will be copied to the index. let's use a shared_ptr here instead.

cpp/src/cylon/indexing/index.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+  }
+  virtual std::shared_ptr<BaseIndex> BuildIndex(std::shared_ptr<arrow::Table> &input_table,
+                                                const int index_column,
+                                                bool index_drop,
+                                                std::shared_ptr<arrow::Table> &output_table) = 0;
+};
+
+template<class ARROW_T, typename CTYPE = typename ARROW_T::c_type>
+class HashIndexKernel : public IndexKernel {
```
nirandaperera, line 133: we should move this class to the cpp

cpp/src/cylon/table.cpp
```diff
@@ -1047,6 +1047,38 @@ Status DistributedUnique(std::shared_ptr<cylon::Table> &in,
   return Unique(shuffle_out, cols, out);
 }
 
+Status Table::Set_Index(const int index_column, bool drop_index, std::shared_ptr<cylon::Table> &out) {
+  std::shared_ptr<arrow::Table> arrow_out;
+  std::shared_ptr<arrow::Table> input_table = table_;
```
nirandaperera, line 1052: we wouldnt need this object right?

cpp/src/cylon/table.cpp
```diff
@@ -1047,6 +1047,38 @@ Status DistributedUnique(std::shared_ptr<cylon::Table> &in,
...
+    const arrow::Result<std::shared_ptr<arrow::Table>> &res = input_table->CombineChunks(cylon::ToArrowPool(ctx));
+    RETURN_CYLON_STATUS_IF_ARROW_FAILED(res.status())
+    input_table = res.ValueOrDie();
+  }
+
+  std::shared_ptr<cylon::IndexKernel> kernel = CreateHashIndexKernel(input_table, index_column);
+  std::shared_ptr<cylon::BaseIndex> bi = kernel->BuildIndex(input_table, index_column, drop_index, arrow_out);
+  table_ = std::move(arrow_out);
+  base_index_ = std::move(bi);
+  return Status();
```
nirandaperera, line 1064: let's not return an empty status. use Status::OK()

cpp/src/cylon/table.hpp
```diff
@@ -199,6 +206,7 @@ class Table {
   std::shared_ptr<arrow::Table> table_;
   bool retain_ = true;
   std::vector<std::shared_ptr<cylon::Column>> columns_;
+  std::shared_ptr<cylon::BaseIndex> base_index_ = nullptr;
```
nirandaperera, line 209: shouldn't we have a default rangeIndex here? other than a nullptr?

nirandaperera (changes requested):

cpp/src/cylon/indexing/index.cpp
vibhatha: Yes, this is just a very initial functional draft. Just want to point out the rough workflow.

PR #287, opened 2021-01-22
op framework refactor 

nirandaperera:


supunkamburugamuve (approved):

supunkamburugamuve (approved):

cpp/src/cylon/arrow/arrow_all_to_all.hpp
```diff
@@ -131,7 +128,7 @@ class ArrowAllToAll : public ReceiveCallback {
                 const std::vector<int> &source,
                 const std::vector<int> &targets,
                 int edgeId,
-                std::shared_ptr<ArrowCallback> callback,
+                ArrowCallback callback,
```
chathurawidanage, line 131: Does it work when the Callback is copied?

chathurawidanage (changes requested):

cpp/src/cylon/arrow/arrow_all_to_all.hpp
nirandaperera: Yes, it is. Since ArrowCallback is a lambda and as long as we pass args by reference (i.e. [&]) it should work. I felt like there is no need for a shared_ptr here because it is just being handed over from the operator level to comms. So, at most, there will only be 2 copies.

PR #294, opened 2021-02-05
Java build fixed

chathurawidanage:


supunkamburugamuve:
@chathurawidanage are we shipping the so with the jar?

chathurawidanage:
> @chathurawidanage are we shipping the so with the jar?

Yes.. libcylon.so and libcylonjni.so are included in the jar...

supunkamburugamuve:
Great. What do you think about linking arrow and glog statically to libcylon and include in the jar?

supunkamburugamuve:
glog is already linked statically. For the java case, we can build cylon with static arrow libraries. or we can put the arrow library inside the jar as well.

chathurawidanage:
A few months back, when I tried to load arrow.so to JVM, it failed. I'll try again to see if it works..

supunkamburugamuve:
In that case, shall we link it statically to cylon for Java?

chathurawidanage:
I tried this approach. But seems we might have to include few other libs including openmpi libs if the jar needs to be standalone.

supunkamburugamuve:
Got you. I think may. be a better approach for us is to decouple distributed from local and provide those as two separate libraries. That will make it easier to ship the local version.

supunkamburugamuve (approved):

PR #297, opened 2021-02-09
Trying a fix for test failure issue 

nirandaperera:
Based on this 
https://stackoverflow.com/questions/47759793/travis-ci-c-build-succeed-but-travis-exited-with-1-and-doesnt-passes-the-test


nirandaperera:
This doesn't seem to be working!

PR #301, opened 2021-02-10
Aggregate addons

vibhatha:


python/pycylon/data/aggregates.pxd
```diff
@@ -24,7 +24,10 @@ cdef extern from "../../../cpp/src/cylon/compute/aggregate_kernels.hpp" namespac
         CCOUNT 'cylon::compute::COUNT'
         CMIN 'cylon::compute::MIN'
         CMAX 'cylon::compute::MAX'
-
+        CMEAN 'cylon::compute::MEAN'
+        CVAR 'cylon::compute::VAR'
+        CNUNIQUE 'cylon::compute::NUNIQUE'
+        CQUANTILE 'cylon::compute::QUANTILE'
```
nirandaperera, line 30: To give quantile we would have to expose the "compute::Op"s. Because we need to get that quantile argument. What shall we do there? @vibhatha

nirandaperera (approved):

python/pycylon/data/aggregates.pxd
vibhatha: Yes, I will work on another PR for this. I also thought about this.

PR #305, opened 2021-02-12
API and Functional Fixes for UNO 

vibhatha:


cpp/src/cylon/indexing/index.cpp
```diff
@@ -75,13 +75,31 @@ cylon::RangeIndex::RangeIndex(int start, int size, int step, arrow::MemoryPool *
...
                                                                                         end_(size),
                                                                                         step_(step) {
+  arrow::Status ar_status;
+  std::vector<int64_t> range_index_values;
+  std::shared_ptr<arrow::Array> index_arr;
+  for (int i = 0; i < end_ - start_; i=i+step_) {
+    range_index_values.push_back(i);
+  }
+  arrow::Int64Builder builder(pool);
+  ar_status = builder.AppendValues(range_index_values);
```
nirandaperera, line 85: since we can calculate the number of values that are there in the final final `index_arr`, we should reserve that num of values in the builder and do an unsafe append. I think using a vector is redundant here.

nirandaperera (changes requested):

nirandaperera (changes requested):

cpp/src/cylon/indexing/index.cpp
```diff
@@ -75,13 +75,29 @@ cylon::RangeIndex::RangeIndex(int start, int size, int step, arrow::MemoryPool *
                                                                                         start_(start),
                                                                                         end_(size),
                                                                                         step_(step) {
+  arrow::Status ar_status;
+  std::shared_ptr<arrow::Array> index_arr;
```
nirandaperera, line 79: let's drop this temp `index_arr` variable and use `index_arr_` class variable

PR #315, opened 2021-02-16
Fixing issues with quicksort by using introsort

supunkamburugamuve:


nirandaperera:
@supunkamburugamuve  shall we add a test case for an already sorted array as well?

supunkamburugamuve:
@nirandaperera there is a test case for an already sorted array

nirandaperera:
> @nirandaperera there is a test case for an already sorted array
@supunkamburugamuve  great! :+1:

nirandaperera (approved):

PR #333, opened 2021-02-27
Concat op

vibhatha:


cpp/src/cylon/join/join.cpp
```diff
@@ -75,6 +75,8 @@ arrow::Status do_inplace_sorted_join(const std::shared_ptr<arrow::Table> &left_t
                                      int64_t left_join_column_idx,
                                      int64_t right_join_column_idx,
                                      cylon::join::config::JoinType join_type,
+                                     std::string left_table_prefix,
```
nirandaperera, line 78: let's pass strings by reference here.

cpp/src/cylon/join/join_config.hpp
```diff
@@ -31,60 +33,82 @@ class JoinConfig {
...
+             int left_column_idx,
+             int right_column_idx,
+             JoinAlgorithm algorithm,
+             std::string left_table_suffix,
+             std::string right_table_suffix)
+      : type(type),
+        algorithm(algorithm),
+        left_column_idx(left_column_idx),
+        right_column_idx(right_column_idx),
+        left_table_suffix(left_table_suffix),
```
nirandaperera, line 59: let's do a `std::move` here. I believe this should be the only copy we make.

cpp/src/cylon/join/join_utils.cpp
```diff
@@ -32,17 +32,19 @@ arrow::Status build_final_table_inplace_index(size_t left_inplace_column, size_t
...
                                               const std::shared_ptr<arrow::Table> &right_tab,
+                                              const std::string &left_table_prefix,
+                                              const std::string &right_table_prefix,
                                               std::shared_ptr<arrow::Table> *final_table,
                                               arrow::MemoryPool *memory_pool) {
   // creating joined schema
   std::vector<std::shared_ptr<arrow::Field>> fields;
   // TODO: get left and right suffixes from user if needed and update it here and replace in the schema with newfileds
-  std::string  prefix = "lt-";
+  std::string  prefix = left_table_prefix;
```
nirandaperera, line 42: we can use a `std::string` pointer here. It's a minor thing.

nirandaperera (changes requested):
In the method signatures, let's use `const std::string&`s for left and right prefixes.

PR #336, opened 2021-03-02
fix #335 adding elementwise arithmaritic ops

### LINKED ISSUE #335: element-wise arithmetic ops are not supported for pycylon tables
ex: 
```python
table:pycylon.Table 
table['c'] = table['a'] + table['b']
```

nirandaperera:
fix #335


vibhatha (commented):
pa.scalar() faster than cast op?

nirandaperera:
> pa.scalar() faster than cast op?

I thought it was neater because ultimately, libarrow accepts an arrow::Scalar. Was there a perf difference there?

vibhatha (approved):

PR #337, opened 2021-03-02
adding standard deviation

nirandaperera:


vibhatha (changes requested):
Should we include a test case for python compared to Pandas?

PR #342, opened 2021-03-03
fix #340 

### LINKED ISSUE #340: Using of macros in macros.hpp evaluates an object twice (if status failed)

nirandaperera:
fix #340 Using of macros in macros.hpp evaluates an object twice (if status failed)

### DIFF
--- a/cpp/src/cylon/util/macros.hpp
+++ b/cpp/src/cylon/util/macros.hpp
@@ -19,25 +19,37 @@
   LOG(ERROR) << msg ; \
   return cylon::Status(code, msg);
 
-#define RETURN_CYLON_STATUS_IF_FAILED(status) \
-  if (!status.is_ok()) { \
-    return status; \
-  };
+#define RETURN_CYLON_STATUS_IF_FAILED(expr) \
+  do{                                       \
+    const auto& _st = (expr);               \
+    if (!_st.is_ok()) {                     \
+      return _st;                           \
+    };                                      \
+  } while (0);
 
-#define LOG_AND_RETURN_CYLON_STATUS_IF_FAILED(status) \
-  if (!status.is_ok()) { \
-    LOG(ERROR) << status.get_msg() ; \
-    return status; \
-  };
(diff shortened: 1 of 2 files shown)


cpp/src/cylon/util/macros.hpp
```diff
@@ -19,25 +19,37 @@
   LOG(ERROR) << msg ; \
   return cylon::Status(code, msg);
 
-#define RETURN_CYLON_STATUS_IF_FAILED(status) \
-  if (!status.is_ok()) { \
-    return status; \
-  };
+#define RETURN_CYLON_STATUS_IF_FAILED(expr) \
+  do{                                       \
+    const auto& _st = (expr);               \
```
supunkamburugamuve, line 24: it is better not to declare this and just use the expr

nirandaperera: you mean remove the macro?

supunkamburugamuve: I meant const auto& _st

nirandaperera: that's how it was previously. But it was evaluating the expression multiple times. That's why I took this approach. 
ex:
```c++

struct Status {
  Status(bool s, std::string msg) : s(s), msg(std::move(msg)) {
    std::cout << "child " << this->msg << std::endl;
  }
  bool is_ok() const{    return s;  }
  std::string test() const{    return msg;  }
  bool s;
  std::string msg;
};

#define TEST11(_st) \
  if (_st.is_ok()) { \
    std::cout << "msg: " << _st.test() << std::endl ; \
    return 0; \
  };

int main() {
 TEST11(Status(true, "ttttt"))
  return 0;
}
```
Here, `Status(true, "ttttt")` expression is evaluated twice. 
```
child ttttt
msg: child ttttt
ttttt
```

supunkamburugamuve: Here the issue is whatever in the parentheses is used to replace the _st in the macro. I think the best solution is not to use macros. Macros are really not a good idea for C++ programs.

PR #343, opened 2021-03-03
Bug fixes

chathurawidanage:


cpp/src/cylon/util/arrow_utils.cpp
```diff
@@ -228,5 +228,15 @@ arrow::Status SampleArray(const std::shared_ptr<arrow::Array> &arr, uint64_t num
   return SampleArray(std::make_shared<arrow::ChunkedArray>(arr), num_samples, out);
 }
 
+std::shared_ptr<arrow::Array> GetChunkOrEmptyArray(
```
chathurawidanage, line 231: @supunkamburugamuve @nirandaperera @vibhatha Would this be a good approach to replace chunk(0)s. This doesn't look nice though. Any suggestions?

vibhatha: Ultimately Sample array can be our column abstraction?

chathurawidanage: Even I thought about that. But can we create an array without a builder?

chathurawidanage: For random gens, how about making them static?

static std::random_device rd;
  static std::mt19937_64 gen(rd());

nirandaperera: > Even I thought about that. But can we create an array without a builder?

I think we can simply create a`arrow::ArrayData` object with length 0 and pass it to an `arrow::Array`.

nirandaperera: > For random gens, how about making them static?
> 
> static std::random_device rd;
> static std::mt19937_64 gen(rd());

It wouldn't be a problem IMO. But I am more inclined to have a function that creates an empty array. 
```c++
template <typename TYPE>
static inline std::shared_ptr<arrow::Array> create_empty_array();
```

chathurawidanage: > I think we can simply create aarrow::ArrayData object with length 0 and pass it to an arrow::Array.

This is not possible. arrow::Array doesn't have public constructors.

nirandaperera: > > I think we can simply create aarrow::ArrayData object with length 0 and pass it to an arrow::Array.
> 
> This is not possible. arrow::Array doesn't have public constructors.

They do I think. 
https://github.com/apache/arrow/blob/8972b1206e09d70e2027458564215ea41d265d92/cpp/src/arrow/array/array_primitive.h#L46

chathurawidanage:
It's not the generic array type, right?

Regards,
Chathura

cpp/src/cylon/util/arrow_utils.cpp
nirandaperera: > It's not the generic array type, right? Regards, Chathura
Yeah. What we can do is something like this. 
```c++
using ARRAY_T = typename arrow::TypeTraits<ARROW_T>::ArrayType;
const auto& empty_array = std::make_shared<ARRAY_T>(arrow::ArrayData::Make(arrow::TypeTraits<ARROW_T>::type_singleton(), 0));
```
Haven't tested this. But I have a feeling that this works. if this doesn't work, let's leave this as it is.

nirandaperera (approved):

PR #353, opened 2021-03-10
Astype perf improvement

vibhatha:


python/test/test_table_properties.py
```diff
@@ -796,4 +796,4 @@ def test_str_astype():
     assert tb_astype_with_dict.index.values.tolist() == tb.index.values.tolist()
 
 
-test_str_astype()
\ No newline at end of file
+test_str_astype()
```
nirandaperera, line 799: this shouldn't be here IMO. test framework would pick methods with "test_*" name, isn't it?

PR #360, opened 2021-03-16
Initial conda build

supunkamburugamuve:


supunkamburugamuve:
Here is how to install for now

conda create -n mypython3 python=3.8
conda install -c supunk06 pycylon

nirandaperera (approved):
this is working for me. but works only on python 3.7 (not 3.8 as in the comment)

PR #370, opened 2021-04-01
fixing gcc8 sort problem

nirandaperera:


cpp/src/cylon/arrow/arrow_kernels.cpp
```diff
@@ -465,14 +465,13 @@ arrow::Status SortIndicesMultiColumns(arrow::MemoryPool *memory_pool,
...
-    int8_t res = 0;
     for (auto const &comp : comparators) {
-      res = comp->compare(idx1, idx2);
+      auto res = comp->compare(idx1, idx2);
       if (res != 0) {
         return res < 0;
       }
     }
-    return res > 0;
+    return false; // if this point is reached, that means every comparison has returned 0! so, equal. i.e. NOT less
```
supunkamburugamuve, line 474: Is this correct? does this work for ascend descent cases? I didn't check the code completely, just checking to make sure

chathurawidanage: I think it works since ArrayIndexComparator handles the asc, desc.

cpp/src/cylon/table.cpp
```diff
@@ -310,6 +310,11 @@ Status Sort(std::shared_ptr<cylon::Table> &table, const std::vector<int32_t> &so
 
 Status Sort(std::shared_ptr<cylon::Table> &table, const std::vector<int32_t> &sort_columns,
             std::shared_ptr<cylon::Table> &out, const std::vector<bool> &sort_direction) {
+  // if single index sort is passed
```
chathurawidanage, line 313: Should we call this function from single index sort instead and remove cylon::util::SortTable?

supunkamburugamuve (approved):

cpp/src/cylon/arrow/arrow_kernels.cpp
nirandaperera: yeah. ascending and descending is managed in the ArrayIndexComparator

cpp/src/cylon/table.cpp
nirandaperera: you mean bring the sort table logic under this method?

PR #371, opened 2021-04-02
Python op improvement

vibhatha:
Math ops and filter ops are supported with both numpy and arrow. 
It can be distinguished by setting the compute engine.


nirandaperera (approved):
@vibhatha I think we'll have to document these configs in the context.

vibhatha:
> @vibhatha I think we'll have to document these configs in the context.

@nirandaperera I agree, let me add python docs for that.

PR #373, opened 2021-04-02
Dataframe API

chathurawidanage:


vibhatha (commented):
@chathurawidanage 

1. Shall we add test cases with assertions comparing results to corresponding pandas ops. 
2.  Are we adding other set operations (intersect, difference)? 
3.  Joins with indexing, if the tables are already indexed, how should we handle the index for the resultant table
     i). reset index and do the join and set index again

chathurawidanage:
>     2. Are we adding other set operations (intersect, difference)?

Dataframes support these ops through the merge() call. But we can add additional functions to support these two ops.

supunkamburugamuve:
@chathurawidanage do we have these functions through the merge call?

chathurawidanage:
Yes. For intersection, it's an inner join.

For difference, we don't support the "indicator" flag of the merge() call which is required to support diff. That might require backend support. 

> indicator(Unsupported) : bool or str, default False
>             If True, adds a column to the output DataFrame called "_merge" with
>             information on the source of each row. The column can be given a different
>             name by providing a string argument. The column will have a Categorical
>             type with the value of "left_only" for observations whose merge key only
>             appears in the left DataFrame, "right_only" for observations
>             whose merge key only appears in the right DataFrame, and "both"
>             if the observation's merge key is found in both DataFrames.

nirandaperera:
> Yes. For intersection, it's an inner join.
> 
> For difference, we don't support the "indicator" flag of the merge() call which is required to support diff. That might require backend support.
> 
> > indicator(Unsupported) : bool or str, default False
> > If True, adds a column to the output DataFrame called "_merge" with
> > information on the source of each row. The column can be given a different
> > name by providing a string argument. The column will have a Categorical
> > type with the value of "left_only" for observations whose merge key only
> > appears in the left DataFrame, "right_only" for observations
> > whose merge key only appears in the right DataFrame, and "both"
> > if the observation's merge key is found in both DataFrames.

This needs an additional filter step, but I think @chathurawidanage is correct. Pandas doesn't have a set difference op directly. 
https://medium.com/@rodwan.bakkar/pandas-set-difference-fde7f4381b53

supunkamburugamuve (approved):

Issue #381, opened 2021-04-07
PyCylon doesn't work with Python 3.7


supunkamburugamuve:
python table_compute_examples.py 
Traceback (most recent call last):
  File "table_compute_examples.py", line 19, in <module>
    import pycylon as cn
  File "/root/anaconda3/envs/myenv/lib/python3.7/site-packages/pycylon/__init__.py", line 26, in <module>
    from pycylon.frame import DataFrame
  File "/root/anaconda3/envs/myenv/lib/python3.7/site-packages/pycylon/frame.py", line 17, in <module>
    from typing import Hashable, List, Dict, Literal, Optional, Sequence, Union, Final
ImportError: cannot import name 'Literal' from 'typing' (/root/anaconda3/envs/myenv/lib/python3.7/typing.py)

vibhatha:
Could this be a Python version issue?

chathurawidanage:
Literal is available in python 3.8+. This should work fine in 3.8+. But
it's just typings, there is no harm in removing it other than a warning
from the IDE.

chathurawidanage:
So should we keep them or remove?

vibhatha:
The test cases are braking if the system has Python 3.7.*.

If the system has 3.7.x (test with the terminal)

My suggestion is to remove since we have sort of agreed to stay with 3.7 to
support proposed work with CuDF, but we can make them as issues to be
replaced
once we migrate to 3.8.

@Supun, @Chathura what do you think?

With Regards,
Vibhatha Abeykoon

supunkamburugamuve:
Yes, lets support Python 3.7. So +1 for removing

PR #385, opened 2021-04-07
Changes to examples

chathurawidanage:


chathurawidanage:
Fixed

https://github.com/cylondata/cylon/issues/383
https://github.com/cylondata/cylon/issues/381

python/examples/dataframe/concat.py
```diff
@@ -19,6 +17,13 @@
 # distributed unique
 env = CylonEnv(config=MPIConfig())
 
-print("Distributed Unique", env.rank)
-df4 = df1.concat(axis=0, objs=[df2, df3], env=env)
-print(df4)
+if env.world_size > 1:
```
supunkamburugamuve, line 20: do we need this check here?

chathurawidanage: Not required. But it would be meaningless to say it as distributed otherwise. This can be removed, if it is not necessary.

supunkamburugamuve: My suggestion is to remove it. Having an unwanted line can complicate things

supunkamburugamuve (approved):

PR #389, opened 2021-04-08
Adding a logging switch to python

chathurawidanage:
Logging will be disabled by default in python-mode. For debugging purposes, if logging needs to be enabled, it can be enabled as follows.

```bash
export CYLON_LOG_LEVEL=<log_level>
python python/examples/dataframe/join.py
```

| Log Level | Flag |
|-----------|------|
| INFO      | 0    |
| WARN      | 1    |
| ERROR     | 2    |
| FATAL     | 3    |

Additionally, this can be done programmatically as follows.

```python
from pycylon.util.logging import log_level, disable_logging


log_level(0) # set an arbitrary log level
disable_logging() # disable logging completely
```


chathurawidanage:
Fixing https://github.com/cylondata/cylon/issues/382

supunkamburugamuve (approved):

PR #412, opened 2021-04-22
adding finalize check and removing destructor finalize call.

vibhatha:


chathurawidanage:
This might break the examples. Shall we explicitly add the finalize call to the examples?

vibhatha:
> This might break the examples. Shall we explicitly add the finalize call to the examples?

added.

nirandaperera (approved):

nirandaperera:
On a separate note, shall we add closeable interface to the env?

vibhatha:
> On a separate note, shall we add closeable interface to the env?

Yes, I think it will be interesting to add that kind of functionality. Let's review this one separately.

PR #420, opened 2021-04-28
Bug fixes in dataframe 

### LINKED ISSUE #415: add SortOptions to df.sort_values(...)

### LINKED ISSUE #416: df.concat uses Table.union instead of Table.concat
This Table.Union would remove duplicates, whereas the concat semantic is to keep duplicates

nirandaperera:
fixes #415 add SortOptions to df.sort_values(...)
fixes #416 df.concat uses Table.union instead of Table.concat

### DIFF
--- a/python/examples/dataframe/concat.py
+++ b/python/examples/dataframe/concat.py
@@ -1,17 +1,28 @@
-from pycylon import DataFrame, CylonEnv
-from pycylon.net import MPIConfig
 import random
 
-df1 = DataFrame([random.sample(range(10, 100), 50),
-                 random.sample(range(10, 100), 50)])
-df2 = DataFrame([random.sample(range(10, 100), 50),
-                 random.sample(range(10, 100), 50)])
-df3 = DataFrame([random.sample(range(10, 100), 50),
-                 random.sample(range(10, 100), 50)])
+import pyarrow as pa
+
+from pycylon import DataFrame, CylonEnv
+from pycylon.net import MPIConfig
+
+t1 = pa.table([random.sample(range(10, 100), 5),
+               random.sample(range(10, 100), 5)], names=['0', '0'])
+
+df1 = DataFrame([random.sample(range(10, 100), 5),
+                 random.sample(range(10, 100), 5)])
+df2 = DataFrame([random.sample(range(10, 100), 5),
+                 random.sample(range(10, 100), 5)])
+df3 = DataFrame([random.sample(range(10, 100), 10),
+                 random.sample(range(10, 100), 10)])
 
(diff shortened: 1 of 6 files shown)


python/examples/dataframe/concat.py
```diff
@@ -1,17 +1,28 @@
...
 from pycylon.net import MPIConfig
-import random
 
-df1 = DataFrame([random.sample(range(10, 100), 50),
-                 random.sample(range(10, 100), 50)])
-df2 = DataFrame([random.sample(range(10, 100), 50),
-                 random.sample(range(10, 100), 50)])
-df3 = DataFrame([random.sample(range(10, 100), 50),
-                 random.sample(range(10, 100), 50)])
+t1 = pa.table([random.sample(range(10, 100), 5),
```
vibhatha, line 8: Is this necessary?

python/pycylon/data/table.pyx
```diff
@@ -2366,7 +2365,7 @@ cdef class Table:
         return PyLocIndexer(self, "iloc")
 
     @staticmethod
-    def concat(tables: List[Table], axis: int = 0, join: str = 'inner', algorithm: str = 'sort', distributed=False):
+    def concat(tables: List[Table], axis: int = 0, join: str = 'inner', algorithm: str = 'sort'):
```
vibhatha, line 2368: A local concat operation is needed even in the distributed setting. 
Shall we add separate methods for local and distributed or we can keep this flag?

vibhatha: Note: This is used as it is in the UNO application

nirandaperera: Oh. I see your point. Ishall we add a `distributed_concat` to the table API (like in other operations)?

vibhatha: Yeah that sounds great :+1:

nirandaperera: great. I'll add that now. And the use it in dataframe api accordingly.

python/pycylon/frame.py
```diff
@@ -1858,6 +1689,7 @@ def sort_values(
             na_position="last",
             ignore_index=False,
             key=None,
+            sort_options: Dict = None,
```
chathurawidanage, line 1692: @supunkamburugamuve

chathurawidanage, line 1692: @nirandaperera Let's change this to accept multiple args instead.

vibhatha (approved):

chathurawidanage (approved):

PR #465, opened 2021-07-20
Add iloc operation to DataFrame

deHasara:


supunkamburugamuve:
@deHasara could you add documentation for iloc operation as well?

deHasara:
I added the documentation.

supunkamburugamuve (approved):

PR #499, opened 2021-09-14
Feature/equal

kaiyingshan:


cpp/src/examples/equal_example.cpp
```diff
@@ -0,0 +1,57 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
```
supunkamburugamuve, line 4: I think this one may not need an example, I think we generally have lot of examples, but my opinion is we should have less examples. may be a better case for this one is a test case. Others may feel differently and I would like to hear.

python/pycylon/pycylon/frame.py
```diff
@@ -363,7 +363,13 @@ def to_table(self) -> cn.Table:
 
     def to_csv(self, path, csv_write_options: CSVWriteOptions):
         self._table.to_csv(path=path, csv_write_options=csv_write_options)
-        
+
+    def equal(self, df: DataFrame):
+        return self._table.equal(df._table)
```
supunkamburugamuve, line 368: Lets have one method (equals) add a parameter for unordered_euqal

cpp/src/cylon/table.cpp
```diff
@@ -988,6 +988,35 @@ Status DistributedUnique(std::shared_ptr<cylon::Table> &in, const std::vector<in
...
+
+  std::shared_ptr<cylon::Table> out_a, out_b;
+  auto status = Sort(const_cast<std::shared_ptr<cylon::Table>&>(a), indices_a, out_a, true);
+  if (!status.is_ok()) {
+      throw "unable to sort table in unordered equal: " + status.get_msg();
+  }
+
+  status = Sort(const_cast<std::shared_ptr<cylon::Table>&>(b), indices_b, out_b, true);
+  if (!status.is_ok()) {
+      throw "unable to sort table in unordered equal: " + status.get_msg();
```
supunkamburugamuve, line 1014: we haven't used exceptions in our code. So we may need to remove these until we come to a consensus on that.

python/pycylon/pycylon/frame.py
```diff
@@ -363,7 +363,13 @@ def to_table(self) -> cn.Table:
...
+
+    def unordered_equal(self, df: DataFrame):
```
supunkamburugamuve, line 370: we need a test case for the equal methods

cpp/src/cylon/table.cpp
```diff
@@ -988,6 +988,27 @@ Status DistributedUnique(std::shared_ptr<cylon::Table> &in, const std::vector<in
   return Unique(shuffle_out, cols, out);
 }
 
+bool Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylon::Table>& b, bool ordered) {
```
supunkamburugamuve, line 991: Lets return a Status and use bool as an out parameter

cpp/src/cylon/table.cpp
```diff
@@ -988,6 +988,27 @@ Status DistributedUnique(std::shared_ptr<cylon::Table> &in, const std::vector<in
...
+    std::vector<int32_t> indices_a(col), indices_b(col);
+    for(int i = 0; i < col; i++) {
+      indices_a[i] = indices_b[i] = i;
+    }
+
+    std::shared_ptr<cylon::Table> out_a, out_b;
+    auto status = Sort(const_cast<std::shared_ptr<cylon::Table>&>(a), indices_a, out_a, true);
+    status = Sort(const_cast<std::shared_ptr<cylon::Table>&>(b), indices_b, out_b, true);
+
+    return out_a->get_table()->Equals(*out_b->get_table().get());
```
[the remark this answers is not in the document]
supunkamburugamuve: I think we have DistributedJoin and Join. Like that may be this can another method called DistributedEqual?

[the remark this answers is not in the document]
kaiyingshan: Thank you for pointing this out!
But could you explain why do we use allReduce instead of just reduce?

cpp/src/cylon/table.hpp
```diff
@@ -425,7 +425,7 @@ Status Unique(std::shared_ptr<cylon::Table> &in, const std::vector<int> &cols,
 Status DistributedUnique(std::shared_ptr<cylon::Table> &in, const std::vector<int> &cols,
 						 std::shared_ptr<cylon::Table> &out);
 
-
+bool Equals(const std::shared_ptr<cylon::Table> &a, const std::shared_ptr<cylon::Table> &b, bool ordered);
```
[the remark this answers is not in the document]
kaiyingshan: I have a question about the function signature. All the other operators return Status, should we return bool (as what we do now) for equal, or should we also return status (and use bool& in parameter for return value)?

cpp/src/cylon/table.cpp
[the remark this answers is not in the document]
kaiyingshan: Another question, if we partition based hash, I think we will lose the order information. Can we just partition based on the index?

supunkamburugamuve:
Unordered Equal
------------------------

Local unordered equal

Subtract(A, B) should give 0 elements and Len(A) == Len(B)

We can implement this in a distributed way using two distributed subtract operations. 

DistributedSubstract(A, B) should give 0 elements and Len(A) == Len(B)

Ordered Equal
--------------------

Local ordered equal

Here we are looking only at the data in a single process. So each process does the following with their own tables A and B

e = (A == B)

Distributed ordered equal

Now after this, if we need to check whether this equality holds in all the processes, we can do AllReduce one to achieve this. 

Here the assumption is that we think the ordering of the elements in the processes is what we need. If someone needs a different ordering they need to do a partition first and then do the equal operator.

supunkamburugamuve:
We are running the same program in all the processes. The processes can distinguish themselves by process ids. Process ids start from 0 and go up to n - 1 where n is the number of parallel processes.

So lets say we execute the following program as an example. Here we assume there are n files with names like file_path0, file_path1 to file_path(n-1).

Table t = ReadCSV(file_path + GetRank())

At runtime each process will read a different file to create a partition. Our number of partitions = parallel processes.

Now you can do the equal operator on each process. 

I think there is one interesting case I missed in the ordered table equal case.

Assume we have 4 processes and our two tables have following number of rows.

         P0    P1     P2       P3
T1    10     8.       12.     10
T2.   5.     17.       8.       10

Now if we take just local equal without doing any shuffling it will say unequal.

To do the order equal, we need to shift the rows of one table to match the other table. For example, we can shift T2 to match T1. Here we assume the rows are ordered from 0th process to n-1 th process. 

P1 will send the first 5 rows to P0 and last 2 rows to P2.  

If we can implement this type of equal with ordering that will be the best.

nirandaperera:
@kaiyingshan this needs a rebase. :slightly_smiling_face:

PR #522, opened 2021-10-13
Adding array flattening 

nirandaperera:
Row-wise flattens a set of arrays to a single Binary array. 
ex: 
```
a1 = [a, b, c, d], a2 = [e, f, g, h] --> [ae, bf, cg, dh]
```

If there are nulls in either of arrays, each element of the output array would be arranged as
follows.
```
|  1byte  | each index 1 byte         |    row size in bytes^^   |
|num_nulls|<...sparse null indices...>|<...flattened row data...>|
 ```
^^ While flattening null values, for fixed sized data (int, float, etc) an empty element (i.e. 0)
will be appended. For variable sized data (str, binary), null values would have empty slots.

ex: 
```
a1 = [200, null, 300, null], a2 = [eee, fff, null, null] --> 
[
|#nulls  |idx    |flattened data|  
|    0   |       |200.eee|
|    1   |0      |0.fff|
|    1   |1      |300.|
|    2   |0.1    |0.|
]
NOTE: characters '|' and '.' are just for clarity.
```

Flattening is only supported for a maximum of 255 arrays at a time.  

Test cases are done by emulating this behavior in python 
https://colab.research.google.com/drive/15laPovDw6R_JwW_vYihR7Z-298X06nyo?usp=sharing

### DIFF
--- a/cpp/src/examples/CMakeLists.txt
+++ b/cpp/src/examples/CMakeLists.txt
@@ -59,7 +59,6 @@ cylon_add_exe(indexing_example)
 cylon_add_exe(sorting_example)
 cylon_add_exe(multicolumn_sorting_example)
 cylon_add_exe(multi_idx_join_example)
-cylon_add_exe(unique_perf)
 
 if (CYLON_PARQUET)
     cylon_add_exe(parquet_union_example)


ahmet-uyar:
[Flattening is only supported for a maximum of 255 arrays at a time.]
In modin paper, they say that they would like to support dataframes with billions of columns. "In many settings, e.g., when dealing with graph adjacency matrices in neuroscience or genomics, the number of rows and number of columns can both be very large."

maybe we can increase 255.

PR #526, opened 2021-10-31
Repartition - CPU

kaiyingshan:
repartition c++ implementation and tests


cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+
+  if(acc != dest_sizes_acc.back() + rows_per_partition.back()) {
+    return Status(
+    cylon::Code::ValueError,
+    "rows_per_partition total number of rows does not align with actual number of rows. Received " +
+        std::to_string(dest_sizes_acc.back() + rows_per_partition.back()) + ", Expected " +
+        std::to_string(acc));
+  }
+
+  std::vector<std::pair<int, int>> send_to = find_mapping(start_idx, num_row, rows_per_partition, dest_sizes_acc);
```
nirandaperera, line 1224: I'd rather use a `vector<int64>` instead of `vector<pair<int, int64>>` here. We can make the index of the vector correspond to the rank, isn't it? :thinking:

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+  uint32_t no_of_partitions = receive_build_rank_order.size();
+  std::vector<uint32_t> outPartitions(num_row);
+
+  int idx = 0;
+  auto itr = outPartitions.begin();
+  for(auto p: send_to) {
+    std::fill(itr + idx, itr + idx + p.second, (uint32_t) receive_build_rank_order[p.first]);
+    idx += p.second;
+  }
```
nirandaperera, line 1233: could you please clarify this loop? I didn't understand the `receive_build_rank_order[p.first]` part

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+
+  std::vector<std::shared_ptr<arrow::Table>> partitioned_tables;
+  RETURN_CYLON_STATUS_IF_FAILED(Split(table, no_of_partitions, outPartitions, partitioned_tables));
+
+  std::shared_ptr<arrow::Schema> schema = table->get_table()->schema();
```
nirandaperera, line 1238:
```suggestion
  const auto& schema = table->get_table()->schema();
```

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+  int world_size = table->GetContext()->GetWorldSize();
+  int num_row = table->Rows();
+  std::vector<int64_t> size = { num_row };
+  std::vector<int64_t> sizes;
+  mpi::AllGather(size, world_size, sizes);
+
+  int total = 0;
+  for(int n: sizes) {
+    total += n;
+  }
```
nirandaperera, line 1275: use `std::accumulate`

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+}
+
+// repartition to `world_size` number of partitions evenly
+Status Repartition(const std::shared_ptr<cylon::Table>& table,
+                   std::shared_ptr<cylon::Table> *output) {
+  int world_size = table->GetContext()->GetWorldSize();
+  int num_row = table->Rows();
+  std::vector<int64_t> size = { num_row };
+  std::vector<int64_t> sizes;
+  mpi::AllGather(size, world_size, sizes);
```
nirandaperera, line 1270: comments on the previous method applies here as well

cpp/test/repartition_test.cpp
```diff
@@ -0,0 +1,259 @@
...
+static void verify_test(std::vector<std::vector<std::string>>& expected, std::shared_ptr<Table>& output) {
+    std::stringstream ss;
+    output->PrintToOStream(ss);
+    std::string s;
+    int i = 0;
+    while(ss>>s) {
+        REQUIRE(s == expected[RANK][i++]);
+    }
+    REQUIRE(i == expected[RANK].size());
+}
```
nirandaperera, line 31: you can use `ARROW_EQUALS` macro here. To create expected arrow table, use `TableFromJSON` in arrow_test_utils header

nirandaperera (changes requested):
@kaiyingshan Great work! I requested some changes. Let me know if those make sense/ or you need any clarification.

ahmet-uyar:
I have implemented this https://github.com/ahmet-uyar/cylon/blob/repartition/cpp/src/cylon/repartition.hpp
Basically there two two methods I think you have also implemented: 
DivideRowsEvenly (this is less significant but, may remove replicated code)
RowIndicesToAll (this calculates the list of indices to send partitions to)

ahmet-uyar:
While testing, I recommend to test with empty tables and partitions also. 
for example: 
```
        std::vector<int64_t> rows_per_partition = {0, 0, 0, 0};
        std::vector<int64_t> rows_per_partition = {12, 0, 0, 0};
        std::vector<int64_t> rows_per_partition = {6, 0, 6, 0};
```
I also recommend to test non-numeric data types such as strings and dates. 
there are data files for that: 

- data/mpiops/sales_nulls_nunascii_x.csv

nirandaperera:
> I have implemented this https://github.com/ahmet-uyar/cylon/blob/repartition/cpp/src/cylon/repartition.hpp Basically there two two methods I think you have also implemented:
> 
>     * DivideRowsEvenly (this is less significant but, may remove replicated code)
> 
>     * RowIndicesToAll (this calculates the list of indices to send partitions to)-

Yes, I also saw that there were some duplication in the 2 PRs. Best would be to merge 1 PR first and then reuse utils from it in the other.

cpp/src/cylon/table.hpp
```diff
@@ -407,6 +407,20 @@ Status Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylo
...
+                   const std::vector<int>& receive_build_rank_order,
+                   std::shared_ptr<Table> *output);
+
+// TODO: change this to reference
+
+Status Repartition(const std::shared_ptr<cylon::Table>& table,
+                   const std::vector<int64_t>& rows_per_partition,
+                   std::shared_ptr<cylon::Table> *output);
+
+Status Repartition(const std::shared_ptr<cylon::Table>& table,
```
ahmet-uyar, line 421: we may remove this function. Previous function can get rows_per_partition with default value of empty vector. If the vector is empty, it can perform even repartitioning.

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+Status Repartition(const std::shared_ptr<cylon::Table>& table, 
+                   const std::vector<int64_t>& rows_per_partition,
+                   const std::vector<int>& receive_build_rank_order,
+                   std::shared_ptr<Table> *output) {
+  int world_size = table->GetContext()->GetWorldSize();
+  int rank = table->GetContext()->GetRank();
+  int num_row = table->Rows();
+  std::vector<int64_t> size = { num_row };
+  std::vector<int64_t> sizes;
+  mpi::AllGather(size, world_size, sizes);
```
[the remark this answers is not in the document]
kaiyingshan: from the code, it seems like mpi::AllGather can only take vectors as arguments, but MPI_Allgather can take pointers. Should I use it?

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+        std::to_string(rows_per_partition.size()) + ", Expected " +
+        std::to_string(world_size));
+  }
+
+  // collecting start index of each process after the repartition
+  std::vector<int64_t> dest_sizes_acc(rows_per_partition.size());
+  dest_sizes_acc[0] = 0;
+  for(int i = 1; i < dest_sizes_acc.size(); i++) {
+    dest_sizes_acc[i] = dest_sizes_acc[i - 1] + rows_per_partition[i - 1];
+  }
```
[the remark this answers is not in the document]
kaiyingshan: I think rows_per_partition is non-modifiable, so we can't do it in-place here

cpp/src/cylon/table.cpp
kaiyingshan: p.first is the partition, p.second is the number of elements.
for one partition, it may need to send, for example, `a1` elements to partition `receive_build_rank_order[0]`, and then `a2` elements to partition `receive_build_rank_order[1]`, then `send_to` will be `{{a1, 0}, {a2, 1}}`.

nirandaperera:
@kaiyingshan I merged #528 now. Ping @ahmet-uyar if you need any help with those utils.

cpp/src/cylon/table.cpp
```diff
@@ -1056,26 +1126,124 @@ Status Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylo
   return Status::OK();
 }
 
+static Status RepartitionToMatchOtherTable(const std::shared_ptr<cylon::Table> &a, const std::shared_ptr<cylon::Table> &b, std::shared_ptr<cylon::Table> * b_out) {
+  int world_size = a->GetContext()->GetWorldSize();
+  int num_row = a->Rows();
+
+  if(num_row == 0) {
+    return Status::OK();
+  }
```
nirandaperera, line 1135: I think since we are returning `Status::OK`, we need to set `b_out`. We can simply assign `a` to it.

cpp/src/cylon/table.cpp
```diff
@@ -1056,26 +1126,124 @@ Status Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylo
...
+
+  std::vector<int64_t> rows_per_partition;
+  RETURN_CYLON_STATUS_IF_FAILED(mpi::AllGather({ num_row }, world_size, rows_per_partition));
+  RETURN_CYLON_STATUS_IF_FAILED(Repartition(b, rows_per_partition, b_out));
+
+  return Status::OK();
```
nirandaperera, line 1141: nit
```suggestion
  return Repartition(b, rows_per_partition, b_out);
```

cpp/src/cylon/table.cpp
```diff
@@ -1056,26 +1126,124 @@ Status Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylo
   return Status::OK();
 }
 
+static Status RepartitionToMatchOtherTable(const std::shared_ptr<cylon::Table> &a, const std::shared_ptr<cylon::Table> &b, std::shared_ptr<cylon::Table> * b_out) {
+  int world_size = a->GetContext()->GetWorldSize();
+  int num_row = a->Rows();
```
nirandaperera, line 1131: Rows are usually int64

cpp/src/cylon/table.cpp
```diff
@@ -1056,26 +1126,124 @@ Status Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylo
...
+                   const std::vector<int64_t>& rows_per_partition,
+                   const std::vector<int>& receive_build_rank_order,
+                   std::shared_ptr<Table> *output) {
+  int world_size = table->GetContext()->GetWorldSize();
+  int rank = table->GetContext()->GetRank();
+  int num_row = table->Rows();
+
+  if(num_row == 0) {
+    return Status::OK();
+  }
```
nirandaperera, line 1182:
```suggestion
  if(num_row == 0) {
  *output = table;
    return Status::OK();
  }
```

cpp/src/cylon/table.cpp
nirandaperera: bump

cpp/src/cylon/table.cpp
```diff
@@ -1076,6 +1153,138 @@ Status DistributedEquals(const std::shared_ptr<cylon::Table> &a, const std::shar
...
+  if (!table->IsRetain()) {
+    const_cast<std::shared_ptr<Table> &>(table).reset();
+  }
+
+  std::shared_ptr<arrow::Table> table_out;
+  RETURN_CYLON_STATUS_IF_FAILED(all_to_all_arrow_tables_preserve_order(table->GetContext(), schema, partitioned_tables, table_out));
+
+  *output = std::make_shared<cylon::Table>(table->GetContext(), table_out);
+
+  return Status::OK();
```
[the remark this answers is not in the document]
nirandaperera: bump

cpp/src/cylon/table.cpp
```diff
@@ -1056,26 +1126,124 @@ Status Equals(const std::shared_ptr<cylon::Table>& a, const std::shared_ptr<cylo
...
+}
+
+// repartition to `world_size` number of partitions evenly
+Status Repartition(const std::shared_ptr<cylon::Table>& table,
+                   std::shared_ptr<cylon::Table> *output) {
+  int world_size = table->GetContext()->GetWorldSize();
+  int num_row = table->Rows();
+  std::vector<int64_t> size = { num_row };
+  std::vector<int64_t> sizes;
+  mpi::AllGather(size, world_size, sizes);
```
nirandaperera, line 1241: Status of this call needs to be checked

cpp/test/repartition_test.cpp
nirandaperera: bump

python/pycylon/test/test_repartition.py
```diff
@@ -0,0 +1,17 @@
...
+>> pytest -q python/pycylon/test/test_repartition.py
+"""
+
+def test_repartition():
+    env=CylonEnv(config=MPIConfig())
+    df1, _ = create_df([random.sample(range(10, 300), 50),
+                            random.sample(range(10, 300), 50),
+                            random.sample(range(10, 300), 50)])
+    df2 = df1.repartition([50], None, env=env)
+    assert_eq(df1, df2)
```
nirandaperera, line 17: to run this test, you need add this under `test/test_all.py`. Otherwise, it wouldn't run. 

And I think this could be extended for multiple workers. 
```suggestion
from utils import create_df,assert_eq
from pycylon.net import MPIConfig
import random

"""
Run test:
>> pytest -q python/pycylon/test/test_repartition.py
"""

def test_repartition():
    env=CylonEnv(config=MPIConfig())
    world_sz = env.get_world_size()  
    df1, _ = create_df([random.sample(range(10, 300), 50),
                            random.sample(range(10, 300), 50),
                            random.sample(range(10, 300), 50)])
    df2 = df1.repartition([50 for _  in range(world_sz)], None, env=env) # distributed repartition 
    assert_eq(df1, df2)  # still the local partitions would be equal 
```

nirandaperera (changes requested):
@kaiyingshan I made some minor comments. I also saw that there were some previous comments that need to be addressed. 
Overall, it looks good to me. Good job @kaiyingshan  :-) 
Let's address these comments and merge this on green CI!

nirandaperera (approved):

nirandaperera:
@kaiyingshan Thank you very much for doing this! Great work.. :+1:

PR #539, opened 2021-12-10
Adding CPU serializer

nirandaperera:


cpp/src/cylon/serialize/table_serialize.cpp
```diff
@@ -0,0 +1,354 @@
...
+    buffer_sizes->push_back((int) ((data.length + 1) * sizeof(int64_t)));
+    data_buffers->push_back(reinterpret_cast<const uint8_t*>(data.GetValues<int64_t>(1)));
+    return Status::OK();
+  }
+
+  return {Code::Invalid, "unsupported offset type for serialization " + type->ToString()};
+}
+
+Status CylonTableSerializer::Make(const std::shared_ptr<Table> &table,
+                                  std::shared_ptr<TableSerializer> *serializer) {
```
ahmet-uyar, line 114: why this is not std::shared_ptr<TableSerializer> &serializer

cpp/src/cylon/serialize/table_serialize.cpp
```diff
@@ -0,0 +1,354 @@
...
+  // the buffers would have to be managed at the allocator level.
+  return SliceMutableBuffer(arrow_buf, offset, size);
+}
+
+Status DeserializeTable(const std::shared_ptr<CylonContext> &ctx,
+                        const std::shared_ptr<arrow::Schema> &schema,
+                        const std::vector<std::shared_ptr<Buffer>> &received_buffers,
+                        const std::vector<int32_t> &buffer_sizes,
+                        const std::vector<int32_t> &buffer_offsets,
+                        std::shared_ptr<Table> *output) {
```
ahmet-uyar, line 263: why not &output

cpp/src/cylon/serialize/table_serialize.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+ private:
+
+  const std::shared_ptr<arrow::Table> table_;
+  const int32_t num_buffers_;
+  const std::vector<int32_t> buffer_sizes_;
+  const std::vector<const uint8_t *> data_buffers_;
+
+  // when there are boolean buffers with non-byte-boundary-offsets, we can not get a byte* to the
+  // data start. So, we'll have to make copy to new buffer. These new buffers will be kept here.
```
ahmet-uyar, line 57: why do we keep new buffer in a different place. can't we just keep the new buffers in data_buffers_?

cpp/src/cylon/util/macros.hpp
```diff
@@ -57,7 +57,7 @@
     if ((arrow_table)->column(0)->num_chunks() > 1){            \
       const auto &res = (arrow_table)->CombineChunks((pool));   \
       RETURN_CYLON_STATUS_IF_ARROW_FAILED(res.status());        \
-      (arrow_table) = res.ValueOrDie();                         \
+      (arrow_table) = std::move(res).ValueOrDie();              \
```
ahmet-uyar, line 60: is there a reason for the parentheses around arrow_table variable?

cpp/test/serialize_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include "common/test_header.hpp"
+
+#include "cylon/serialize/table_serialize.hpp"
+#include "cylon/arrow/arrow_buffer.hpp"
+
+namespace cylon {
+namespace test {
+
+TEST_CASE("serialize table", "[serialization]") {
```
ahmet-uyar, line 23: are we testing empty tables?

cpp/test/sync_comms_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "common/test_header.hpp"
+
+namespace cylon {
+namespace test {
+
+TEST_CASE("all gather table", "[sync comms]") {
```
ahmet-uyar, line 20: we should test empty tables: all empty tables, some empty tables.
Single row tables with nulls.

cpp/src/cylon/arrow/arrow_buffer.cpp
```diff
@@ -0,0 +1,45 @@
...
+ArrowAllocator::~ArrowAllocator() = default;
+
+Status ArrowAllocator::Allocate(int64_t length, std::shared_ptr<Buffer> *buffer) {
+  CYLON_ASSIGN_OR_RAISE(auto alloc_buf, arrow::AllocateBuffer(length, pool))
+  *buffer = std::make_shared<ArrowBuffer>(std::move(alloc_buf));
+  return Status::OK();
+}
+
+int64_t ArrowBuffer::GetLength() const {
+  return 0;
```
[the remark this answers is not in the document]
nirandaperera: that's a huge mistake! thanks..

cpp/src/cylon/arrow/arrow_types.cpp
```diff
@@ -153,5 +153,34 @@ std::shared_ptr<DataType> ToCylonType(const std::shared_ptr<arrow::DataType> &ar
   return nullptr;
 }
 
+Type::type ToCylonTypeId(const std::shared_ptr<arrow::DataType> &type) {
```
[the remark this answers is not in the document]
nirandaperera: there's no real difference because we are passing `const &`. We haven't used shared_ptr deref operator that much. That's why I didn't use it.

cpp/src/cylon/net/communicator.hpp
```diff
@@ -36,6 +38,24 @@ class Communicator {
 
   virtual ~Communicator() = default;
 };
+
+class SyncCommunicator {
```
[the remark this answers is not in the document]
nirandaperera: I am thinking to add something like, `cylon::Column` (holds `arrow::Array`) and `cylon::Scalar` (holds `arrow::Scalar`), and add those collective ops. We can leave arrow::Buffer counterparts in mpi_operator.hpp IMO.

cpp/src/cylon/serialize/table_serialize.cpp
nirandaperera: This is because of the google style guide. We initially followed it, but then decided to use mutable references for output params. I wanted to reverse it (again) mainly because of [this](https://stackoverflow.com/a/26441753/4116268) reason TBH.

cpp/src/cylon/serialize/table_serialize.hpp
nirandaperera: `data_buffers_` are just pointers right? 
Actually these are the copied bool buffers. There wouldn't be any tables that holds these `extra_buffers_`. They would only be created during the serializing step, and the serializer owns those buffers (untill they are consumed).

cpp/src/cylon/util/macros.hpp
nirandaperera: yeah, that a macros thing. It means to replace `(arrow_table)` site with the passed `arrow_table` param. (so you can pass something like `auto foo`)

cpp/src/cylon/serialize/table_serialize.hpp
ahmet-uyar: ok, so the class member may be needed for deallocating the memory. 
But, we may delete the method extra_buffers() if we already point to this buffer from data_buffers_.

cpp/src/cylon/serialize/table_serialize.cpp
ahmet-uyar: I kind of like using & in the method signature and avoid using & when calling. But I get the point.

cpp/src/cylon/serialize/table_serialize.hpp
nirandaperera: well, I havent added a method to clear them (yet). These would be deleted when the serializer gets destroyed.
If we are shuffling we dont have to worry about it because, its already copied to a remote memory location by the time the serializer falls off.

ahmet-uyar (approved):

Issue #554, opened 2021-12-23
[Bug] unordered equal gives wrong result


nirandaperera:
I was trying the following and the assertion fails. Looks like the tables have the same values but not in the same order. I was expecting the unordered equals to capture this.
```python 
def test_parquet_join():
    cdf1 = DataFrame(read_table('data/input/parquet1_0.parquet'))
    cdf2 = DataFrame(read_table('data/input/parquet2_0.parquet'))
    expected = DataFrame(read_table('data/output/join_inner_1_0.parquet'))

    out = cdf1.merge(cdf2, how='inner', on=[0], algorithm='sort', suffixes=('lt-', 'rt-'))

    print(expected)
    print(out)
    assert(expected.equals(out, ordered=False))  # should pass 
```

output 
```
python/pycylon/test/test_parquet.py::test_parquet_join FAILED            [100%]   lt-0   lt-1  rt-0   rt-1
0     3  0.025     3  0.025
1    26  0.676    26  0.394
2    26  0.394    26  0.394
3    29  0.291    29  0.979
4    29  0.291    29  0.979
5    29  0.291    29  0.979
6    29  0.291    29  0.979
7    57  0.153    57  0.314
8    57  0.314    57  0.314
9    61  0.685    61  0.685
   lt-0   lt-1  rt-0   rt-1
0     3  0.025     3  0.025
1    26  0.394    26  0.394
2    26  0.676    26  0.394
3    29  0.291    29  0.979
4    29  0.291    29  0.979
5    29  0.291    29  0.979
6    29  0.291    29  0.979
7    57  0.153    57  0.314
8    57  0.314    57  0.314
9    61  0.685    61  0.685

test/test_parquet.py:31 (test_parquet_join)
def test_parquet_join():
        cdf1 = DataFrame(read_table('data/input/parquet1_0.parquet'))
        cdf2 = DataFrame(read_table('data/input/parquet2_0.parquet'))
        expected = DataFrame(read_table('data/output/join_inner_1_0.parquet'))
    
        out = cdf1.merge(cdf2, how='inner', on=[0], algorithm='sort', suffixes=('lt-', 'rt-'))
    
        print(expected)
        print(out)
    
>       assert(expected.equals(out, ordered=False))  # should pass
E       assert False
[...]

PR #558, opened 2021-12-30
Fixing 554 

nirandaperera:
testing #554


nirandaperera:
@kaiyingshan it seems to be failing inside conda env. This is quite strange. :thinking:

nirandaperera:
@kaiyingshan could you pleasw review this PR?

cpp/src/cylon/util/arrow_utils.cpp
```diff
@@ -376,5 +376,16 @@ arrow::Status MakeEmptyArrowTable(const std::shared_ptr<arrow::Schema> &schema,
...
 }
 
+bool CheckArrowTableContainsChunks(const std::shared_ptr<arrow::Table> &table,
+                                   const std::vector<int> &columns) {
+  if (columns.empty()) {
+    return std::any_of(table->columns().begin(), table->columns().end(),
+                       [](const auto &col) { return col->num_chunks() > 1; });
+  } else {
+    return std::any_of(columns.begin(), columns.end(),
+                       [&](const auto &i) { return table->column(i)->num_chunks() > 1; });
```
kaiyingshan, line 386: I think it would be better to use `int` (or `auto`) here
```suggestion
                       [&](int i) { return table->column(i)->num_chunks() > 1; });
```

python/pycylon/test/test_parquet.py
```diff
@@ -36,5 +36,5 @@ def test_parquet_join():
 
     out = cdf1.merge(cdf2, how='inner', on=[0], algorithm='sort', suffixes=('lt-', 'rt-'))
 
-    # assert(expected.equals(out, ordered=False))  # should pass
+    assert(expected.equals(out, ordered=False))  # should pass
```
kaiyingshan, line 39: I think we can remove this comment now
```suggestion
    assert(expected.equals(out, ordered=False))
```

PR #574, opened 2022-02-28
Dist sort cpu

kaiyingshan:
implement distributed sort regular sampling, change `distributedSort` default option to regular sampling (involves changing SortOptions)


cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace test {
+TEST_CASE("Dist sort testing", "[dist sort]") {
+    std::string path1 = "../data/input/csv1_" + std::to_string(RANK) + ".csv";
+    std::shared_ptr<Table> table1, table2;
+
+    auto read_options = io::config::CSVReadOptions().UseThreads(false);
+
+    CHECK_CYLON_STATUS(FromCSV(ctx, std::vector<std::string>{path1},
+                            std::vector<std::shared_ptr<Table> *>{&table1},
+                            read_options));
```
nirandaperera, line 29:
```suggestion
    CHECK_CYLON_STATUS(FromCSV(ctx, path1,&table1, read_options));
```

cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    SECTION("dist_sort_test_1") {
+        std::shared_ptr<Table> out, out2, repartitioned;
```
nirandaperera, line 32:
```suggestion
        std::shared_ptr<Table> out, out2;
```

cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+        std::shared_ptr<arrow::Table> arrow_output;
+
+        auto before = time(NULL); // TODO: test performance & eveness of row distribution
+        auto status = DistributedSort(table1, {0, 1}, out, {1, 1});
+        auto after = time(NULL);
+        
+        if(RANK == 0) {
+        out->Print();
+        }
+        DistributedSort(table1, {0, 1}, out2, {1, 1}, {0, 0, SortOptions::INITIAL_SAMPLE});
```
nirandaperera, line 43: we'll check reference as well!
```suggestion
        CHECK_CYLON_STATUS(DistributedSort(table1, {0, 1}, out2, {1, 1}, {0, 0, SortOptions::INITIAL_SAMPLE}));
```

cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+        std::shared_ptr<Table> out, out2, repartitioned;
+        auto ctx = table1->GetContext();
+        std::shared_ptr<arrow::Table> arrow_output;
+        auto status = DistributedSort(table1, {1, 0}, out, {0, 0});
+        DistributedSort(table1, {1, 0}, out2, {0, 0}, {0, 0, SortOptions::INITIAL_SAMPLE});
+        REQUIRE(status.is_ok());
+        bool eq;
+        status = DistributedEquals(out, out2, eq);
+        REQUIRE(eq);
+    }
```
nirandaperera, line 72: there's a lot of duplicated code in these sections. why don't you add a separate `run_dist_sort_test(...)` method, and call separate test cases with the sections?

cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+*/
+
+#include "common/test_header.hpp"
+#include "test_utils.hpp"
+#include <cylon/table.hpp>
+
+namespace cylon {
+namespace test {
+TEST_CASE("Dist sort testing", "[dist sort]") {
+    std::string path1 = "../data/input/csv1_" + std::to_string(RANK) + ".csv";
```
nirandaperera, line 22: Rather than using files, this is an alternative method to generate a global table within a test case and slice it based on the rank. 
https://github.com/cylondata/cylon/blob/b2c0820744e148b807da3f44658bb91c64cc8271/cpp/test/aggregate_test.cpp#L319-L409

These tests could be done in the following way (This removes the dependency on INITIAL_SAMPLE sorting method for tests) 
1. Slice global table to create each partition for ranks
2. Do distributed sort
3. At rank 0, locally sort the global table using `arrow::compute::SortIndices` + `arrow::compute::Take`. Other ranks would have an empty table with the same schema. 
4. Call ordered `DistributedEquals` on globally sorted table created by step 3 and distributed sort result

nirandaperera (changes requested):
@kaiyingshan Thank you for doing this. I noticed a couple of things. 
1. Please use a code formatter (we use Google style and wrapping at 100 chars) - Almost all IDEs provide this support
2. We could reuse more of the comparators and util methods. I've added comments where things could be simplified using our existing methods. 

Please ping me if you need any clarification.

cpp/src/cylon/table.cpp
```diff
@@ -427,16 +497,298 @@ Status Sort(const std::shared_ptr<Table> &table, const std::vector<int32_t> &sor
...
+Status GetSplitPoints(std::shared_ptr<Table>& sample_result,
+                      const std::vector<int>& sort_columns,
+                      const std::vector<bool>& sort_orders,
+                      int num_split_points,
+                      std::shared_ptr<Table>& split_points) {
+  auto ctx = sample_result->GetContext();
+
+  std::vector<std::shared_ptr<cylon::Table>> gather_results;
+  // net::MPICommunicator comm;
+  RETURN_CYLON_STATUS_IF_FAILED(ctx->GetCommunicator()->AllGather(sample_result, &gather_results));
```
ahmet-uyar, line 636: i think, we need to gather all sample tables to the first worker. We don't need allgather here.

cpp/src/cylon/table.cpp
```diff
@@ -427,16 +497,298 @@ Status Sort(const std::shared_ptr<Table> &table, const std::vector<int32_t> &sor
...
+
+  int num_split_points = std::min(merged_table->Rows(), (int64_t)ctx->GetWorldSize() - 1);
+
+  RETURN_CYLON_STATUS_IF_FAILED(SampleTableUniform(merged_table, num_split_points, split_points, ctx));
+
+  return Status::OK();
+}
+
+Status GetSplitPoints(std::shared_ptr<Table>& sample_result,
+                      const std::vector<int>& sort_columns,
```
ahmet-uyar, line 628: when sample_result only has sort columns, we should not need sort_columns here.

cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,98 @@
+/*
```
ahmet-uyar, line 1: it would be nice to add separate test cases for important methods such as tableBinarySearch, MergeSortedTable.

ahmet-uyar (changes requested):

cpp/src/cylon/table.cpp
```diff
@@ -427,16 +497,298 @@ Status Sort(const std::shared_ptr<Table> &table, const std::vector<int32_t> &sor
...
+                        const std::vector<bool>& sort_orders,
+                        std::shared_ptr<Table>& out) {
+  std::shared_ptr<Table> concatenated;
+  std::vector<int64_t> table_indices(tables.size()), table_end_indices(tables.size());
+  int acc = 0;
+  for(int i = 0; i < table_indices.size(); i++) {
+    table_indices[i] = acc;
+    acc += tables[i]->Rows();
+    table_end_indices[i] = acc;
+  }
```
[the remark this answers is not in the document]
kaiyingshan: > By looking at the values of the vectors here, I think you can use a single `tables.size() + 1` vector for this job! then the i'th `table_end_indices` would be `table_indices[i+1]`

I intentionally used 2 tables, because the `table_indices` is used to mark the next element to be merged in this table; values in `table_indices` will be changed and compared to value in `table_end_indices`.

cpp/src/cylon/table.cpp
```diff
@@ -427,16 +497,298 @@ Status Sort(const std::shared_ptr<Table> &table, const std::vector<int32_t> &sor
...
+    temp_v.push_back(table_indices[t]);
+    filter.UnsafeAppend(table_indices[t]);
+    table_indices[t] += 1;
+    if(table_indices[t] < table_end_indices[t]) {
+      pq.push(t);
+    }
+  }
+  
+  CYLON_ASSIGN_OR_RAISE(auto take_arr, filter.Finish());
+  CYLON_ASSIGN_OR_RAISE(auto take_res, (arrow::compute::Take(concatenated->get_table(), take_arr)));
```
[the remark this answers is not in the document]
kaiyingshan: The sub-arrays are not sorted in initial sampling, but they are sorted in regular sampling. As result, we can use a merge in regular sampling, which is n log(k) in time (n = num items, k = world size). In initial sampling, we need to sort, which is n log(n) in time.

Do you think using merge will benefit us (e.g. more performant)? Or is it actually slower because of more overhead?

cpp/src/cylon/table.cpp
```diff
@@ -427,16 +497,298 @@ Status Sort(const std::shared_ptr<Table> &table, const std::vector<int32_t> &sor
...
+  std::vector<std::unique_ptr<cylon::DualArrayIndexComparator>> comparators(sort_columns.size());
+  for(int64_t i = 0; i < sort_columns.size(); i++) {
+    std::unique_ptr<cylon::DualArrayIndexComparator> comp;
+    RETURN_CYLON_STATUS_IF_FAILED(CreateDualArrayIndexComparator(
+      cylon::util::GetChunkOrEmptyArray(arrow_sorted_table->column(sort_columns[i]), 0), 
+      cylon::util::GetChunkOrEmptyArray(arrow_split_points->column(sort_columns[i]), 0), &comp, sort_order[i]
+    ));
+
+    comparators[i] = std::move(comp);
+  }
```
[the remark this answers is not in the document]
kaiyingshan: How exactly do we specify `sort_columns` and `sort_order`? It seems like the `Make` method only takes one vector for each table, and I think they are `sort_columns`
https://github.com/cylondata/cylon/blob/3344bf950765302492bd9da00536526c33159d3b/cpp/src/cylon/arrow/arrow_comparator.hpp#L277-L281

ahmet-uyar (approved):

cpp/src/cylon/table.cpp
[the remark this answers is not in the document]
nirandaperera: I felt like the end result is the same if we use prefix sum.

cpp/src/cylon/table.cpp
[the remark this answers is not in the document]
nirandaperera: (we already discussed this offline)

cpp/test/dist_sort_test.cpp
```diff
@@ -0,0 +1,287 @@
...
+    out2 = std::make_shared<Table>(ctx, arrow_empty_table);
+  }
+  std::shared_ptr<Table> out3;
+  status = Repartition(out2, &out3);
+  status = DistributedEquals(out3, out, eq);
+  REQUIRE(eq);
+}
+
+namespace test {
+TEST_CASE("Dist sort testing", "[dist sort]") {
```
nirandaperera, line 51: Let's parameterize this test case (let's parameterize field `a` for `ArrowNumericTypes`). 

Take a look at this 
https://github.com/cylondata/cylon/blob/2759a303b5f9d601229bc5c8321586e06f072495/cpp/test/sync_comms_test.cpp#L231
This is Catch2 docs on templated testcases
https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#type-parametrised-test-cases

nirandaperera, line 51: Let's parameterize this test case (let's parameterize field `a` for `ArrowNumericTypes`). 

Take a look at this 
https://github.com/cylondata/cylon/blob/2759a303b5f9d601229bc5c8321586e06f072495/cpp/test/sync_comms_test.cpp#L231
This is Catch2 docs on templated testcases
https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#type-parametrised-test-cases

nirandaperera (changes requested):
@kaiyingshan I made some comments with 1 possible bug. Other that this looks good to me. Let me know if you need anything from my end.

nirandaperera:
conda has introduced some `-external` artifacts for OpenMPI 4.1.3* and it is messing up the entire conda build.

PR #579, opened 2022-04-02
Adding UCC build

nirandaperera:


vibhatha:
Thanks for this PR @nirandaperera 
I will look into this one.

cpp/src/examples/CMakeLists.txt
```diff
@@ -42,7 +42,20 @@ macro(cylon_add_exe EXENAME)
...
+#        if (CYLON_UCC)
+#            target_link_libraries(${EXENAME} ucc)
+#            target_link_libraries(${EXENAME} ucc /home/niranda/git/ucc/install/lib/ucc/libucc_cl_basic.so
+#                    /home/niranda/git/ucc/install/lib/ucc/libucc_cl_hier.so
+#                    /home/niranda/git/ucc/install/lib/ucc/libucc_mc_cpu.so
+#                    /home/niranda/git/ucc/install/lib/ucc/libucc_tl_ucp.so)
+##            target_include_directories(${EXENAME} INTERFACE ${UCC_INCLUDE_DIRS})
+#            include_directories(${UCC_INCLUDE_DIRS})
+#        endif (CYLON_UCC)
+#    endif (CYLON_UCX)
```
vibhatha, line 58: Do we need these?

nirandaperera:
I believe there's an issue in the `ucc_example.cpp` linking. I was able to compile and run the example based on this 
https://github.com/openucx/ucc/wiki/UCC-Allreduce-example

But I still couldn't get it properly linked inside cylon examples unfortunately. :-(

cpp/src/examples/ucc_example.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+#define STR(x) #x
+#define UCC_CHECK(_call)                                            \
+    if (UCC_OK != (_call)) {                                        \
+        fprintf(stderr, "*** UCC TEST FAIL: %s\n", STR(_call));     \
+        MPI_Abort(MPI_COMM_WORLD, -1);                              \
+    }
+
+static ucc_status_t oob_allgather(void *sbuf, void *rbuf, size_t msglen,
+                                  void *coll_info, void **req) {
```
vibhatha, line 31: Are we going to C style for method naming or should we go for Camel case?

cpp/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@
 # For more information about CMake, see http://www.cmake.org             #
 ##########################################################################
 cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+#set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```
vibhatha, line 25: Do we need this?

cpp/src/examples/ucc_example.cpp
```diff
@@ -0,0 +1,179 @@
...
+ *  limitations under the License.
+ */
+
+#include <mpi.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <ucc/api/ucc.h>
+#include <iostream>
```
vibhatha, line 21: order of imports

vibhatha (changes requested):
This is just a quick peek on the PR, I will review again. 

Just a thought, should we add some docs on how to download, compile, which versions we are supporting?

vibhatha:
> I believe there's an issue in the `ucc_example.cpp` linking. I was able to compile and run the example based on this https://github.com/openucx/ucc/wiki/UCC-Allreduce-example
> 
> But I still couldn't get it properly linked inside cylon examples unfortunately. :-(

Let's see, I will also look into this.

cpp/CMakeLists.txt
nirandaperera: may be this was the problem. I added this to run examples without having to update `LD_LIBRARY_PATH`. It seems like we'll have to figure out an alternative to that

nirandaperera: I figured out the issue. It turned out that I had a built ucc with a different ucx version than the one in my conda env.

cpp/src/examples/ucc_example.cpp
nirandaperera: This is just an example. We can remove this once UCC is properly integrated.

PR #643, opened 2023-02-06
Added buildable instructions for Rivanna

arupcsedu:
Signed-off-by: Arup Sarker <arupcsedu@gmail.com>

### DIFF
--- a/python/pycylon/test/test_data_split.py
+++ b/python/pycylon/test/test_data_split.py
@@ -61,4 +61,8 @@ def test_data_split():
     env.barrier()
     
     if env.rank == 0:
-        os.remove(data_file)
+        try:
+            os.remove(data_file)
+            return False
+        except OSError:    # file is in use
+            return True
--- a/rivanna/README.md
+++ b/rivanna/README.md
@@ -11,6 +11,13 @@ The solution is to use the guide we provide here.
 
 ## Intsall instructions
 
+Clone the repository
+
+```shell
+git clone https://github.com/cylondata/cylon.git
+cd cylon
+```
+
 There are two ways you can build cylon on Rivanna
 
 ### Fix all dependencies for Cylon on Rivanna without the dependencies of the system gcc and openmpi version.
@@ -91,6 +98,11 @@ which python gcc g++
 python build.py --cpp --test --python --pytest
 
 ```
+Run the Slurm script
+
+```shell
+sbatch rivana/job.slurm
+```
 
 ### Build Cylon by using the loaded module of openmpi and gcc
 
@@ -162,7 +174,10 @@ python build.py --cpp --test --python --pytest
 
 The two version of slurm scripts are created in this folder. Run any of them with the below command:
 
+Run the Slurm script
+
 ```shell
-sbatch job.slurm
+sbatch rivana/job_with_module.slurm
 ```
+
  
--- a/rivanna/README.md
+++ b/rivanna/README.md
@@ -1,0 +1,168 @@
+# Running Cylon on Rivanna
+
+Arup Sarker (arupcsedu@gmail.com, djy8hg@virginia.edu)
+
+## Issues on Summit
+
+Running cylon on rivanna requires a special configuration. We have verified
+that installations based on conda with virtual environment
+
+The solution is to use the guide we provide here.
+
+## Intsall instructions
+
+There are two ways you can build cylon on Rivanna
+
+### Fix all dependencies for Cylon on Rivanna without the dependencies of the system gcc and openmpi version.
+
+
+
+1. Change in yml for conda build: conda/environments/cylon.yml
+Add this libraries
+  - gcc
+  - gxx
+  - gxx_linux-64
+
+
+```shell
+name: cylon_rivanna
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - python=3.10
+  - cmake>=3.23.1,!=3.25.0
+  - arrow-cpp=9
+  - pyarrow=9.0.0
+  - glog
+  - openmpi=4.1.3=ha1ae619_105
+  - ucx>=1.12.1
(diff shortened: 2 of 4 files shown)


python/pycylon/test/test_data_split.py
```diff
@@ -61,4 +61,8 @@ def test_data_split():
     env.barrier()
     
     if env.rank == 0:
-        os.remove(data_file)
+        try:
+            os.remove(data_file)
+            return False
+        except OSError:    # file is in use
+            return True
```
vibhatha, line 68: I assume you're trying to remove the data after doing the test? If that is the case, it is better to leverage `pytest` and use a temporary directory which gets cleaned up by the test framework itself. Looking into the exising code, previous please can also be cleaned up. 

https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

rivanna/README.md
```diff
@@ -0,0 +1,183 @@
+# Running Cylon on Rivanna
+
+Arup Sarker (arupcsedu@gmail.com, djy8hg@virginia.edu)
```
vibhatha, line 3: Generally don't put author names unless it is a blog post or an article. But I am not very much informed with the recent practices, so I could be wrong. :slightly_smiling_face:

rivanna/README.md
```diff
@@ -0,0 +1,183 @@
...
+  - fsspec>=0.6.0
+  - setuptools
+  # they are not needed for using pygcylon or compiling it
+  - pytest
+  - pytest-mpi
+  - mpi4py
+  - gcc
+  - gxx
+  - gxx_linux-64
+
```
vibhatha, line 59: @arupcsedu just a suggestion, would it be nice to keep this in a `requirements.txt` and just mention the usage in the `README.md`. This way, we can keep minimal edits to the markdown file but add dependencies to the `requirements.txt`.

rivanna/README.md
```diff
@@ -0,0 +1,183 @@
...
+  - gcc
+  - gxx
+  - gxx_linux-64
+  
+ ```
+
+Added all blue packages to overcome the dependencies on the preloaded modules.
+
+```shell
+conda env create -f conda/environments/cylon.yml
```
vibhatha, line 65: Should we specify a `Python` version for this?

rivanna/README.md
vibhatha: And if users has additional dependencies, they can create `requirements_extras.txt` to include their own dependencies but use our `requirements.txt` as the base dependencies.

vibhatha:
@arupcsedu Thanks for working on this. I have externally looked at this.  Sorry about my lack of knowledge about the context of the PR. It would be better to include a good description in the PR to give some clarity on what the feature is introducing. And later on when we run into issues, we could rely on this description and debug. 

I assume this is a platform or compute resource where we want to setup Cylon. Please correct me if I am wrong.

Overall the PR looks good to me in external shape, but I haven't and don't know how to test the instructions to see if they function properly. 

Is there a way to include a CI to Cylon to run this with the said platform/resources? If not, it would be better to keep an isolated/internal workload which can test this. 

Would it be wise to include a docker or a singularity image which could be deployed quite easily?

laszewsk:
* docker is not available in rivanna
* my experience from using docker to singularity to rivanna is miserable due to very slow upload tmes.
* singularity has very limited images available on rivanna from my experience it is best tho have a modified requirements such as the one discussed here.
* As the setup Is so complex it is advisable to distribute this setup so that it can also serve as example for others.
* Maybe you want give all platforms a special directory as i imagine this could be an issue for others.
* one issue for adaption of cylon is platform differences. please remember we have a similar issue for summit.

The conclusion I have is that each platform needs a special documentation so that other do not spend hours, days or weeks for the deployment and they can learn from each other

rivanna/README.md
laszewsk: I think its better to not have a requirements_extra.txt but a 
requirements-rivanna.txt
The reason is that on other platforms you may have different versions of non python libraries as default, and you want to make sure that default software provided by the rivanna system staff can be used by having an explicit rivanna requirements.txt.
Also as this is conda, should it not be in something like
conda/environmenst/rivanna/cylon.yaml
?

python/pycylon/test/test_data_split.py
arupcsedu: os.remove(data_file) was used before and caused an intermittent error during running on a windows machine. 

I just added the code with exception handling. This issue is happening only in this test case. So, we should not make it standard for all other test cases.

rivanna/README.md
arupcsedu: I followed this convention by taking the sample from Summit compatibility documents. 
https://github.com/cylondata/cylon/tree/main/summit#readme

rivanna/README.md
arupcsedu: @vibhatha, This is the readme file and I tried to make minimal changes to the existing default build yaml file.
This are the additional default packages:   
- gcc
- gxx
- gxx_linux-64
are needed on top of existing packages.

It is not ensured that these packages will be the same after 2 years.  During that time, there can build errors. 
Moreover, I have found another way to make Cylon compatible with Rivanna. So, developers have to maintain it and might get confused.  We tried to make it as simple as possible. That's why after discussing with @nirandaperera , we have decided to make an additional readme file.

rivanna/README.md
arupcsedu: We don't need it. Rivanna's default python version works perfectly.

rivanna/README.md
arupcsedu: @laszewsk , we thought to create a different yaml file for Rivanna. But we have two solutions and both solution needs minimal different changes on the default cylon yaml file. So, we kept both options open for developers.

nirandaperera (approved):

PR #688, opened 2024-04-18
[Cylon] Arrow 14.0.2 Integration

arupcsedu:
1. Add support for C++17
2. Update dependent library for conda
3. Update compatibility issue within arrow.


cylon-manager:
@arupcsedu why not use arrow 15.0.x ?

UPDATE: never mind. I think latest cudf is at 14.x.x. So, this should be good

arupcsedu:
cudf, pytorch default release version is compatible with 14.0.2

laszewsk:
v 15 is a major new release, so i think you want to use 15

https://arrow.apache.org/release/15.0.0.html

laszewsk:
is it that cudf, pytorch are not compatible with 15?

arupcsedu:
right, Gregor.
We should align with the released cudf and pytorch version.
Arrow 14.0.2 comes with the released package of cudf and pytorch.

Despite this, it will not take that much time to upgrade to Arrow 15 after
this upgrade.
This one is the toughest one.

Best regards
Arup Sarker

mstaylor:
Arup,

Here are some initial suggestions:
1) Move the rivanna directory and update the necessary scripts under target
2) Change the conda environments to use cylon_dev as a name
3) The build completes successfully but there are many warnings that need to be cleaned up.  We should have a build process that emits a minimal amount of warning output from the compiler

I'm testing with UCX now and therefore more suggestions may be forthcoming.

arupcsedu:
I will do it. Thanks

